Issue #54, opened 2022-08-17
`useKeyboardAnimation` hook is not respecting to keyboard size changes


kirillzyusko:
**Describe the bug**

Animated values from `useKeyboardAnimation` are not reacting to keyboard size changes.

**Repo for reproducing**

Just open any screen (`Animated transition`, `Chat`) in example app and try to change the input mode (for example switch keyboard to emoji selection).

**To Reproduce**
1. Go to `Animated transition` screen;
2. Click on `TextInput`
3. See that red and green circles have been moved accordingly.
4. Change keyboard type to emoji selection
5. See that red and green circles hasn't changed its position.

**Expected behavior**

Returned values from `useKeyboardAnimation` should respect to keyboard size changes and should update its values.

**Screenshots**

|Actual|Expected|
|------|----------|
|<img width="207" alt="image" src="https://user-images.githubusercontent.com/22820318/184610556-fac41777-c268-4af1-b3f8-0e11a3ca1117.png">|<img width="204" alt="image" src="https://user-images.githubusercontent.com/22820318/184610479-1b375d0c-ebeb-4cc5-84bd-78165a4abc92.png">|

> Using standard keyboard both circles are located above the keyboard, but when you change keyboard type - red circle (driven by `useKeyboardAnimation`) is getting overlapped by the keyboard.

**Smartphone:**
 - Desktop OS: Mac OS 12.5
 - Device: Pixel 3 (emulator)
 - OS: Android API 32
 - JS engine: JSC
 - Library version: 1.0.1

PR #58, opened 2022-08-20
[RNKC-054] - detect keyboard size changes

kirillzyusko:
## Description

Added `onApplyWindowInsets` into `WindowInsetsAnimationCompat.Callback` to detect keyboard size changes.

## Motivation and Context

It seems like it's only one option to detect such changes. For me it's a little bit strange, that `onProgress`/`onEnd` callback is not called and we need to write additional code to track such cases.

Fixes https://github.com/kirillzyusko/react-native-keyboard-controller/issues/54

### JS
- changed `Events` example - show `height` in all events;

### Android
- fixed a bug with missing `height` in all `keyboardDid*` events;
- detect keyboard size changes;
- renamed class that implements `WindowInsetsAnimationCompat.Callback` to `KeyboardAnimationCallback`;

## How Has This Been Tested?

Tested manually on:
- Xiaomi Redmi Note 5 Pro (Android 9)
- Pixel 3 (Android 12)

## Screenshots (if appropriate):

|Before|After|
|-------|----|
|image|image|

## Checklist

- [x] CI successfully passed

### DIFF
--- a/android/src/main/java/com/reactnativekeyboardcontroller/KeyboardControllerViewManager.kt
+++ b/android/src/main/java/com/reactnativekeyboardcontroller/KeyboardControllerViewManager.kt
@@ -29,9 +29,6 @@ class KeyboardControllerViewManager(reactContext: ReactApplicationContext) : Rea
       return view
     }
 
-    val window = activity.window
-    val decorView = window.decorView
-
     val callback = KeyboardAnimationCallback(
       view = view,
       persistentInsetTypes = WindowInsetsCompat.Type.systemBars(),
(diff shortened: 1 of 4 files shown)


android/src/main/java/com/reactnativekeyboardcontroller/KeyboardControllerViewManager.kt
```diff
@@ -32,6 +32,17 @@ class KeyboardControllerViewManager(reactContext: ReactApplicationContext) : Rea
...
+      view = view,
+      persistentInsetTypes = WindowInsetsCompat.Type.systemBars(),
+      deferredInsetTypes = WindowInsetsCompat.Type.ime(),
+      // We explicitly allow dispatch to continue down to binding.messageHolder's
+      // child views, so that step 2.5 below receives the call
+      dispatchMode = WindowInsetsAnimationCompat.Callback.DISPATCH_MODE_CONTINUE_ON_SUBTREE,
+      context = mReactContext
+    )
+    ViewCompat.setWindowInsetsAnimationCallback(decorView, callback)
+    ViewCompat.setOnApplyWindowInsetsListener(decorView, callback)
```
kirillzyusko, line 45: Many libraries may rely on `decorView` and setup their own handlers (for example `react-native-screens` when `translucent` status bar prop is set). So the handler may be overwritten

Maybe we can setup listeners on our view?

PR #64, opened 2022-09-09
[RNKC-051] - fabric support

kirillzyusko:
## Description

Added basic support for new (fabric) architecture.

## Motivation and Context

The reason of prioritising a fabric adoption is an ability to perform sync communication between Native and JS thread. Especially it may become handy during the implementation of interactive keyboard on Android.

Current implementation allows to use both (old - paper, and new - fabric) architectures.

Unfortunately, it seems like Swift is not supported in new arch, so I had to rewrite some files to ObjC. I hope at some point of time we will bring it back to Swift.

> For fabric adoption I used this [repo](https://github.com/software-mansion-labs/appjs-2022-workshops-fabric) as an example of how to add a basic support.

### CI
- added `cpplint` as C++ codebase most likely will be increased over development iterations;

### JS
- which exposes `isFabricEnabled` and `isTurboModuleEnabled` variables;
- added View/Modules specification for new arch;

### iOS
- changed files extension from `.m` to `.mm` to support C++;
- added fabric support;

### Android
- added fabric support;

## How Has This Been Tested?

Tested manually on:
- iPhone 11 (iOS 15.5, simulator);
- Pixel 3 (API 32, emulator).

## Checklist

- [x] CI successfully passed


src/KeyboardControllerViewNativeComponent.ts
```diff
@@ -0,0 +1,23 @@
...
+// @ts-expect-error - no type definition
+import type { ViewProps } from 'react-native/Libraries/Components/View/ViewPropTypes';
+import type {
+  DirectEventHandler,
+  Float,
+  Int32,
+} from 'react-native/Libraries/Types/CodegenTypes';
+import codegenNativeComponent from 'react-native/Libraries/Utilities/codegenNativeComponent';
+
+type KeyboardMoveEvent = Readonly<{
```
kirillzyusko, line 11: I think similar data structure is already defined as `NativeEvent` in `native.ts`. Would it be possible to use one source of truth?

kirillzyusko: Moving type declaration into separate file will cause an error during pods installation 🤷‍♂️

PR #167, opened 2023-06-06
fix: check if `jsThread` is not null

kirillzyusko:
## 📜 Description

Added a check for `_jsThread==null` before sending event. Should prevent crash on startup in certain conditions.

## 💡 Motivation and Context

image

### iOS
- check `_jsThread` to exist before sending event;

Will be tested in live 🙃 

## 📝 Checklist

- [x] CI successfully passed


kirillzyusko:
Confirmed by @hirbod - it fixed the problem.

PR #173, opened 2023-06-15
feat: duration

kirillzyusko:
## 📜 Description

Added `duration` to plain and worklet events.

## 💡 Motivation and Context

As part of enhancement events metadata it would be good to have `duration` prop. It may allow to create kind of parallax effect (when animation duration is equal, but easing functions are different).

### JS
- added `duration` to spec file;
- updated example app;

### iOS
- extract duration and send it in eventemitter and in view events;

### Android
- extract duration and send it in eventemitter and in view events;

Tested manually on:
- Pixel 7 Pro (Android 13);
- Xiaomi Redmi Note 5 Pro (Android 9);
- iPhone 14 Pro (iOS 16.5);

## 📝 Checklist

- [x] CI successfully passed

### DIFF
--- a/ios/KeyboardMovementObserver.swift
+++ b/ios/KeyboardMovementObserver.swift
@@ -11,7 +11,7 @@ import Foundation
 @objc(KeyboardMovementObserver)
 public class KeyboardMovementObserver: NSObject {
   // class members
-  var onEvent: (NSString, NSNumber, NSNumber) -> Void
+  var onEvent: (NSString, NSNumber, NSNumber, NSNumber) -> Void
   var onNotify: (String, Any) -> Void
   // progress tracker
   private var _keyboardView: UIView?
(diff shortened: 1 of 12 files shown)


ios/KeyboardMovementObserver.swift
```diff
@@ -138,7 +140,7 @@ public class KeyboardMovementObserver: NSObject {
         return
       }
 
-      onEvent("onKeyboardMoveInteractive", position as NSNumber, position / CGFloat(keyboardHeight) as NSNumber)
+      onEvent("onKeyboardMoveInteractive", position as NSNumber, position / CGFloat(keyboardHeight) as NSNumber, 0)
```
kirillzyusko, line 143: Should be `-1` to match Android behavior 🙂

PR #187, opened 2023-06-27
fix: casting UIView to UIView always succeeds

hirbod:
## 📜 Description

The line `result = hostView as? UIView` will not give any error but it is superfluous.

You're traversing a hierarchy of subviews, which are all UIView objects. In Swift, subviews is an array of UIView.

## 💡 Motivation and Context

Just housekeeping a bit :)

### JS
- 
- 

### iOS
- 
- 

### Android
- 
- 

iOS Device and Simulator

## 📝 Checklist

- [ ] CI successfully passed


kirillzyusko:
Thanks @hirbod 
I wanted to remove it for long long time, but forgot about it every time 😅

kirillzyusko (approved):

Issue #185, opened 2023-06-27
can not scroll content when `interpolator` set to `ios`


kirillzyusko:
**Describe the bug**

UI looks jumpy when I'm trying to scroll the content

**Repo for reproducing**

It's reproducible in example app.

**To Reproduce**

1. Go to 'Interactive keyboard (android)' screen
2. Click on 'linear' in top right corner (it'll switch interpolator to ios mode)
3. Try to scroll content as you typically do on iOS
4. keyboard tries to be closed, but then appears and UI looks jumpy

**Expected behavior**

Keyboard shouldn't try to close itself, since gesture occurs in non keyboard area (gesture not affects keyboard position).

**Screenshots**

Uploading telegram-cloud-document-2-5249207706597732183.mp4…

**Smartphone (please complete the following information):**
 - Desktop OS: [e.g. Windows 10, MacOS 10.15.5]
 - Device: Pixel 7 Pro
 - OS: Android 13
 - RN version: 0.71.11
 - RN architecture: old
 - JS engine: Hermes
 - Library version: 1.5.7

**Additional context**

I think we should just call `animateToFinish` only when keyboard position was changed after the gesture. Otherwise we shouldn't call this method.

PR #189, opened 2023-07-08
fix: pass `velocityY=null` if keyboard didn't change its height

kirillzyusko:
## 📜 Description

Don't hide keyboard if gesture didn't affect its position.

## 💡 Motivation and Context

Initially I fixed this problem by excluding a call to `animateToFinish`. However it brought more problems:
- since `InteractiveKeyboard.isInteractive` value couldn't be reset in some cases (when interaction with keyboard was completed), it could lead to unpredictable results (for example `onInteractive` event will be emitted instead of `onMove` for plain keyboard events);
- when you hide a keyboard by pressing hide button when interactive gestures were finished - `onMove` events were not coming (because `progress` was `Infinity` and couldn't be serialised via bridge - it happened because in `onEnd` callback we were updating `persistentKeyboardHeight` value, but `getCurrentKeyboardHeight()` was returning `0` - it happened because `InteractiveKeyboardProvider.shown` had incorrect values).

To overcome issue 1 I decided to call `animateToFinish(null)` if keyboard size wasn't changed. In this case we specify `InteractiveKeyboardProvider.isInteractive = false` and `isInteractive` state becomes correct one.

To overcome issue 2 I had few ideas:
- do not update `persistentKeyboardHeight` in `onEnd` - it fixes the problem, but since we update `isVisible` value in `onEnd` -> state machine becomes broken, because `isVisible` will be `false`, and as a result handler in `onApplyWindowInsets` will not be called;
- update `InteractiveKeyboardProvider.shown` in `KeyboardAnimationController` - it fully fixes the problem (actually we always should update `InteractiveKeyboardProvider.shown` before a call to `controller.finish(boolean)`).

Fixes https://github.com/kirillzyusko/react-native-keyboard-controller/issues/185

### Android
- pass `velocityY=null` if keyboard size wasn't changed;
- update `InteractiveKeyboard.shown` before call `controller.finish(boolean)`

Tested on Pixel 7 Pro.

|Before|After|
|-------|-----|
|<video src="https://github.com/kirillzyusko/react-native-keyboard-controller/assets/22820318/fe4c01d6-b82d-42e9-a4cb-f3ce619edebf">|<video src="https://github.com/kirillzyusko/react-native-keyboard-controller/assets/22820318/567c509b-22c8-4bf5-872d-9273b59a2e13">|

## 📝 Checklist

- [x] CI successfully passed

### DIFF
--- a/android/src/main/java/com/reactnativekeyboardcontroller/views/KeyboardGestureAreaReactViewGroup.kt
+++ b/android/src/main/java/com/reactnativekeyboardcontroller/views/KeyboardGestureAreaReactViewGroup.kt
@@ -31,6 +31,7 @@ class KeyboardGestureAreaReactViewGroup(private val reactContext: ThemedReactCon
   private var lastTouchX = 0f
   private var lastTouchY = 0f
   private var lastWindowY = 0
+  private var keyboardHeight = 0
 
   // react props
   private var interpolator: Interpolator = LinearInterpolator()
@@ -87,6 +88,9 @@ class KeyboardGestureAreaReactViewGroup(private val reactContext: ThemedReactCon
 
         if (isHandling) {
           if (controller.isInsetAnimationInProgress()) {
+            if (keyboardHeight == 0) {
+              this.keyboardHeight = controller.getCurrentKeyboardHeight()
+            }
             // If we currently have control, we can update the IME insets to 'scroll'
             // the IME in
             val moveBy = this.interpolator.interpolate(dy.roundToInt(), this.getWindowHeight() - event.rawY.toInt(), controller.getCurrentKeyboardHeight())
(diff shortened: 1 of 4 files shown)


android/src/main/java/com/reactnativekeyboardcontroller/views/KeyboardGestureAreaReactViewGroup.kt
```diff
@@ -122,10 +126,18 @@ class KeyboardGestureAreaReactViewGroup(private val reactContext: ThemedReactCon
         // Calculate the current velocityY, over 500 milliseconds
         velocityTracker?.computeCurrentVelocity(500)
         val velocityY = velocityTracker?.yVelocity
+        val isKeyboardPositionChanged =
+          // check `isInsetAnimationInProgress()` before, since direct usage of `getCurrentKeyboardHeight()`
+          // may throw exception
+          !controller.isInsetAnimationInProgress() ||
+            this.keyboardHeight != controller.getCurrentKeyboardHeight()
+        // if keyboard was changed after finger movement -> we need to calculate final position
```
kirillzyusko, line 134: if keyboard height was changed

kirillzyusko:
Try to:
- set interpolator="ios"
- scroll to top
- change keyboard size

Hook will not report new keyboard size 😔

PR #204, opened 2023-08-12
docs: use `.lottie` instead of `.json` for faster page loading

kirillzyusko:
### JS
- 
- 

### iOS
- 
- 

### Android
- 
- 

|Before|After|
|------|-----|
|||

## 📝 Checklist

- [ ] CI successfully passed


kirillzyusko:
The lottie file size was decreased, but `Lighthouse` score has significantly dropped. Check:

|Before|After|
|-------|----|
|image|image|

We'll need to find out a root cause of such performance drop.

kirillzyusko:
Closed in favour of https://github.com/kirillzyusko/react-native-keyboard-controller/pull/205

It seems like right now dotlottie always fetches data from URL (and there is no option to cache result). Because of that I always see content jump (or if fix via css anyway I'll see empty space for fraction of seconds which gives non pleasant user experience).

For now I'm postponing experiments with dotlottie, but I may return to it later, when the package will be more mature and will allow to consume `.lottie` files from js bundle to deliver seamless loading experience.

Issue #51, opened 2022-08-12
[RNKC-053] - native-stack + statusBarTranslucent causes incorrect positioning


kirillzyusko:
**Describe the bug**

using a combination of RN 0.69, native-stack and `statusBarTranslucent` causes some issues. Seems like keyboard insets are also applied automatically, and because of that positioning of views is incorrect.

**Repo for reproducing**

https://github.com/kirillzyusko/react-native-keyboard-controller/pull/50

**To Reproduce**
1. Go to Native Stack
2. Click on TextInput

**Expected behavior**

Views shouldn't react on keyboard appearance if they don't subscribed to `useKeyboardAnimation` animated values.

**Screenshots**

|Expected|Actual|
|---------|-------|
|<img width="255" alt="image" src="https://user-images.githubusercontent.com/22820318/184424127-5e5912f8-acc8-4135-8d75-844797cbaeb1.png">|<img width="260" alt="image" src="https://user-images.githubusercontent.com/22820318/184427263-44435502-58df-471f-b749-0d9b7f7b868b.png">|

**Smartphone (please complete the following information):**
 - Desktop OS: Mac OS 12.5
 - Device: Pixel 3 (emulator)
 - OS: Android API 32
 - JS engine: JSC
 - Library version: 1.0.1

**Additional context**

Seems like it wasn't an issue in RN 0.68

kirillzyusko:
The progress can be tracked here: https://github.com/kirillzyusko/react-native-screens/tree/refactor/removed-deprecated-replace-system-window-insets

kirillzyusko:
Just found a workaround: if you don't change soft-input method in screens, where status bar translucency is controlled by `react-native-screens`, then everything works good (comparing to [original](https://github.com/kirillzyusko/react-native-keyboard-controller/blob/cb5fab5b223668efbfe96248516e83754405edd4/src/hooks.ts#L19) hook - here we are not using `useResizeMode` hook):

```ts
const useKeyboardAnimationInNativeStack = () => {
  const context = useContext(KeyboardContext);

  return context.animated;
};
```

This is a temporary workaround - I'm still looking for a complete solution/fix 👀

PR #207, opened 2023-08-17
refactor: new way for handling insets

kirillzyusko:
## 📜 Description

Setup `onApplyWindowInsetsListener` that manages `systemBars` and `navigationBar` on `rootView` to get correct and constant insets.

## 💡 Motivation and Context

If we setup `onApplyWindowInsetsListener` on `EdgeToEdgeReactViewGroup`, then sometimes insets are including keyboard size (it's is `0, 1225` (top, bottom) instead of expected `144, 84`).

I discovered that if we setup `onApplyWindowInsetsListener` on `rootVIew` - in this case insets are always stable. So I reworked this part (worth to note that `onApplyWindowInsetsListener` that detects keyboard size changes still should listen to `EdgeToEdgeReactViewGroup` because otherwise it's not working).

It fixes a problem with content jump, however the old approach (setting `padding` for `action_bar_root`) is not working anymore - instead we should apply margin. Last, but not least - we should setup it asynchronously, because if we setup it in `onAttachedToWindow`/`init` then it'll add undesired padding:

<p align="center">
  image
</p>
<p align="center">
<i>Area in blue - undesired margin</i>
</p>

Fixes https://github.com/kirillzyusko/react-native-keyboard-controller/issues/51

### Android
- apply `margin` to `action_bar_root` instead of `padding`;
- for managing paddings setup `onApplyWindowInsetsListener` on `rootView`, keep `onApplyWindowInsetsListener` on `EdgeToEdgeView` just for detecting keyboard size changes;
- setup `onApplyWindowInsetsListener` on `rootView` asynchronously (to avoid unnecessary margin on mount);
- added `ThemedReactContext` extension that returns `rootView`;

Tested (fabric/paper) on:
- Pixel 7 Pro (Android 13);
- Redmi note 5 Pro (Android 9);

|Before|After|
|------|-----|
|telegram-cloud-photo-size-2-5402402906965133775-y|telegram-cloud-photo-size-2-5402402906965133774-y|

## 📝 Checklist

- [x] CI successfully passed

### DIFF
--- a/android/src/main/java/com/reactnativekeyboardcontroller/KeyboardAnimationCallback.kt
+++ b/android/src/main/java/com/reactnativekeyboardcontroller/KeyboardAnimationCallback.kt
@@ -31,7 +31,6 @@ class KeyboardAnimationCallback(
   val deferredInsetTypes: Int,
   dispatchMode: Int = DISPATCH_MODE_STOP,
   val context: ThemedReactContext?,
-  val onApplyWindowInsetsListener: OnApplyWindowInsetsListener,
 ) : WindowInsetsAnimationCompat.Callback(dispatchMode), OnApplyWindowInsetsListener {
   private var persistentKeyboardHeight = 0.0
   private var isKeyboardVisible = false
@@ -158,7 +157,7 @@ class KeyboardAnimationCallback(
       this.persistentKeyboardHeight = keyboardHeight
     }
 
-    return onApplyWindowInsetsListener.onApplyWindowInsets(v, insets)
+    return insets
   }
 
   override fun onStart(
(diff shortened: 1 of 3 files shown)


thespacemanatee:
I don't think we should set the `margin` here because it will push all the content down which kind of defeats the purpose of setting the status bar to translucent, but correct me if I'm wrong because I didn't really investigate this. I believe this is causing a regression where all my content is being pushed below the status bar even when set to translucent

kirillzyusko:
@thespacemanatee yes, most likely this PR caused your regression. Let's move all our conversation to the issue that you opened 👀

Issue #198, opened 2023-08-08
ease migration


gaultierq:
This lib looks very promising ! Thanks for tackling this live long problem

One drawback imo, I find it difficult to start using it in a large app, relying heavily on different keyboard avoiding views / scroll views.
I found this repo when I needed to animate my modal size. On android I was missing an keyboard event, making it difficult to calculate the available space for my modal.
Opting in this lib gave me the missing event I needed, but also breaks every other keyboard avoiding screens.

I tried to using only `KeyboardEvents`, but it doesn't work without the `KeyboardProvider` context.

I also tried nesting only my modal with the context, and it works once, but once my modal has been displayed, other keyboard avoiding views are not working anymore.

My question is: can we "activate" react-native-keyboard-controller only for some views / screens ? activate it on demand ?

kirillzyusko:
Hello @gaultierq 

I thought about adding `.setEnabled(boolean)` method, but didn't have time for investigation of how it can work/which new issues it can add 🤷‍♂️ 

Right now when you wrap your app in `KeyboardProvider` app starts to behave in the same way as iOS does - it doesn't resize the screen automatically and delegate control fully to you.

If you were relying on `adjustResize`/`adjustPan` behavior, then as a temporary solution you can write your own component that will mimicrate default Android behavior:

```tsx
const AndroidKeyboardResize = ({children}) => {
  const fullHeightRef = useRef(0);
  const [height, setHeight] = useState<undefined | number>(undefined);

  const onLayout = useCallback(({nativeEvent}) => {
    // save view size on component mount
    if (fullHeightRef.current === 0) {
      fullHeightRef.current = nativeEvent.height;
    }
  });

  useEffect(() => {
    const show = KeyboardEvents.addListener("keyboardWillShow", (e) => {
      if (Platform.OS === 'android') {
        // resize screen
        setHeight(fullHeightRef.current - e.height)
      }
    });
    const hide = KeyboardEvents.addListener("keyboardWillHide", (e) => {
      if (Platform.OS === 'android') {
        // restore full height
        setHeight(undefined);
      }
    });
  
    return () => {
      hide.remove();
      show.remove();
    };
  }, []);

  return (
    <View onLayout={onLayout} style={{height}}>
      {children}
    </View>
  );
};
```

Not sure this code works, but I think you can get the idea 😄 And of course let me know what do you think about it 😊

cwhenderson20:
[...]

PR #227, opened 2023-08-29
feat: enabled

kirillzyusko:
## 📜 Description

Added `useKeyboardController` hook that returns `setEnabled` + `enabled` (which allows to disable and enable module on demand).

## 💡 Motivation and Context

Fixes https://github.com/kirillzyusko/react-native-keyboard-controller/issues/233 https://github.com/kirillzyusko/react-native-keyboard-controller/issues/198

### JS
- added `useKeyboardController` hook that gives an access to `setEnabled` method and `enabled` variable (that indicates current state);
- added `enabled` property to specs;
- manage `enabled` property via state (from react) and passing methods via context to change the value of this variable;
- apply and revert monkey patch depending on `enabled` value.

### iOS
- paper: added property setter that calls `mount`/`unmount` depends on the value that comes;
- fabric: detect prop changes in `updateProps` method and call `mount`/`unmount` depends on the value that comes;

### Android
- go to `edge-to-edge` when prop is set (before we did it in `onAttachedToWindow` method);
- enable `edge-to-edge` mode if `onAttachedToWindow` is called more than one time;

Tested on:
- Pixel 3a (Android 13, emulator)
- Pixel 6 Pro (Android 9, emulator)
- Pixel 7 Pro (Android 13, real device);
- Xiaomi Redmi Note 5 Pro (Android 9, real device);
- iPhone 14 Pro (iOS 16.4, iOS 17, simulator);
- iPhone 6S (iOS 15.6, real device).

|Android|iOS|
|--------|---|
|<video src="https://github.com/kirillzyusko/react-native-keyboard-controller/assets/22820318/6fcece94-8252-4fc8-9ab5-6d9b946168e7">|<video src="https://github.com/kirillzyusko/react-native-keyboard-controller/assets/22820318/9e91d79f-cf42-4074-aeed-0f063896e500">|

## 📝 Checklist

- [x] CI successfully passed


src/animated.tsx
```diff
@@ -61,6 +62,7 @@ export const KeyboardProvider = ({
   statusBarTranslucent,
   navigationBarTranslucent,
 }: KeyboardProviderProps) => {
+  const [enabled, setEnabled] = useState(true);
```
avery-discord, line 65: Can you add a prop to `KeyboardProvider` that makes `false` the initial state? This will useful for large apps that want to adopt this lib incrementally.

avery-discord (changes requested):

src/animated.tsx
kirillzyusko: Hello @avery-discord 👋 

I can not make it `false` by default, because it'll break a backward compatibility with the previous versions of the library.

However I can add a property to `KeyboardProvider` like `initiallyEnabled` (which will be `true` by default but you can override it) and this piece of the code will be:

```ts
const [enabled, setEnabled] = useState(initiallyEnabled);
```

Will it work for you?

avery-discord: Yes, that's exactly what I was thinking. Thank you @kirillzyusko !

Issue #154, opened 2023-05-17
Add Simple component for keyboard avoiding


beqramo:
**Is your feature request related to a problem? Please describe.**
Hi,
In most of the cases in app we just want to move button on top if it is open or something like that.
So it is missing base component that will do this thing. it isn't handy to build animation again and again for simply placing button on top.

Because of that, I would suggest adding base Component that will do simple avoiding like the below:
```tsx
import React from 'react';
import { useReanimatedKeyboardAnimation } from 'react-native-keyboard-controller';
import Animated,  { interpolate, useAnimatedStyle, Extrapolation }  from 'react-native-reanimated';
import { useSafeAreaInsets } from 'react-native-safe-area-context';

interface IKeyboardAwareComponent {
  children: React.ReactNode;
  withBottomInset: boolean;
}


function KeyboardAwareComponent({ children, withBottomInset, ...rest }: IKeyboardAwareComponent) {
  const { height } = useReanimatedKeyboardAnimation();
  const { bottom } = useSafeAreaInsets();

  const styles = useAnimatedStyle(() => {
    const paddingBottom = withBottomInset ? interpolate(
      height.value,
      [0, bottom],
      [bottom, 0],
      Extrapolation.IDENTITY,
    ) : 0;
    
    return {
      transform: [{ translateY: height.value }],
      paddingBottom,
    };
  }, [bottom]);

  return (
    <Animated.View style={styles} {...rest}>
      {children}
    </Animated.View>
  );
}

export default KeyboardAwareComponent;

```
This solution will work only if it is a fixed button, without any scroll.
Depending on where you will put the relative position in the parent component it most likely will work in most cases.
Sorry I didn't answer other paragraphs and I removed it.

Thank you.
Great library btw.

beqramo:
@kirillzyusko I updated the codes. there were some stupid mistakes :).
Replicate in what term? For the example above component will do the same thing as `KeyboardAvoidingView` exactly. but as `KeyboardAvoidingView`  doesn't work always, this will work

beqramo:
Oh you mean to add as an improvement, moving the input field with the keyboard if it is going to be behind?

beqramo:
Yes, this lib is amazing. I am indicating some missing components that instead of in the example file can be inside the package itself.
I dropped one of the simplest component that can be exported from the lib.
Nothig special is written above just instead of everyone to create custom component of base the functionality everyone can use exported component from this lib

[...]

PR #229, opened 2023-09-07
feat: components

kirillzyusko:
## 📜 Description

Added `KeyboardAvoidingView` based on `useKeyboardHandler` hook.

## 💡 Motivation and Context

It was requested functionality and I agree that similar components should be provided out-of-box. 

As a first component I decided to re-implement `KeyboardAvoidingView`, because the implementation is very easy 👀 

Later I'm planning to add more components, so stay tuned 😎 

Closes https://github.com/kirillzyusko/react-native-keyboard-controller/issues/154

### JS
- added `KeyboardAvoidingView`;
- added corresponding example app.

### Docs
- added section about `KeyboardAvoidingView`;
- added lottie that shows visual difference between implementations.

Tested on:
- Xiaomi Redmi Note 5 Pro;
- Pixel 7 Pro;
- iPhone 14 Pro (iOS 16.5, iOS 17) <- simulator;
- iPhone 6s (iOS 15.6).

https://github.com/kirillzyusko/react-native-keyboard-controller/assets/22820318/0fc849d4-4c15-41ab-875f-f49c555256f3

## 📝 Checklist

- [x] CI successfully passed


src/components/KeyboardAvoidingView/index.tsx
```diff
@@ -0,0 +1,145 @@
...
+        </View>
+      );
+    }
+
+    return (
+      <Reanimated.View
+        // @ts-expect-error because `ref` from reanimated is not compatible with react-native
+        ref={ref}
+        onLayout={onLayout}
+        style={enabled ? [style, animatedStyle] : [style]}
```
IvanIhnatsiuk, line 136:
```suggestion
        style={enabled ? [style, animatedStyle] : style}
```

src/components/KeyboardAvoidingView/index.tsx
```diff
@@ -0,0 +1,145 @@
...
+        screenHeight - keyboard.heightWhenOpened.value - keyboardVerticalOffset;
+
+      return Math.max(frame.value.y + frame.value.height - keyboardY, 0);
+    }, [screenHeight, keyboardVerticalOffset]);
+
+    const onLayout = useCallback<NonNullable<ViewProps['onLayout']>>(
+      (e) => {
+        if (initialFrame.value === null) {
+          initialFrame.value = e.nativeEvent.layout;
+        }
```
IvanIhnatsiuk, line 83: Could you please advise me what happens if I change the height of the container? I think initialFrame.value will have the initial values and it may cause some bugs. Maybe it would be better to recalculate them every change the size?

src/components/KeyboardAvoidingView/index.tsx
```diff
@@ -0,0 +1,145 @@
...
+          };
+
+        case 'padding':
+          return { paddingBottom: bottomHeight };
+
+        default:
+          return {};
+      }
+    }, [behavior, enabled, relativeKeyboardHeight]);
+
```
IvanIhnatsiuk, line 120: It would be better to create memoized styles to avoid unnecessary re-renders. 

```suggestion
const combinedStyle = useMemo(() => [animatedStyle, style], [animatedStyle, style]);
```

src/components/KeyboardAvoidingView/index.tsx
kirillzyusko: @IvanIhnatsiuk yes, I know about it. The layout should be updated when keyboard is not shown. In current implementation if I remove `if`-statement above - this function will be called every frame when keyboard starts movement, and calculation will be messed up.

I'm going to improve and re-work this piece of code, but in other PR - the current implementation has some edge cases which I'm going to fix a little bit later and I'd like to have a chain of PRs for better tracing of history of changes 😊 (the current PR already big enough and in most of the cases it should work identically to RN variant).

Issue #156, opened 2023-05-23
Example with FlatList


psteinroe:
**Is your feature request related to a problem? Please describe.**
We are building a chat with your (awesome) package, and it works very smoothly. Thanks! However, at some point, we are experiencing performance issues due to the usage of ScrollView. Is FlatList also supported somehow?

**Describe the solution you'd like**
Add example using FlatList.

**Describe alternatives you've considered**
Using ScrollView, but the performance is bad.

kirillzyusko:
@psteinroe What is the issues with FlatList do you experience?

psteinroe:
> @psteinroe What is the issues with FlatList do you experience?

That was a swift response! ❤️ I am having troubles getting a chat to work. In your examples, you are inverting views, add fakeViews etc. All this is not really possible FlashList since it uses `renderItem` and does not take children.

kirillzyusko:
@psteinroe for adding `fakeView` you can use `ListFooterComponent`/`ListHeaderComponent`. Inverting should work in the same way (i. e. just apply styles to parent view and content style).

psteinroe:
that does not work for some reason. It results in some weird and indeterministic behaviours. It seems like applying transform to `contentContainerStyle` does not really work properly since any rotation is lost after navigating. Do you have any ideas? Otherwise, I will put this on hold, and try to create PR with a FlatList example to reproduce.

kirillzyusko:
@psteinroe I don't have any ideas. This rotation was used to make `ScrollView` to scroll from bottom to top - you also can achieve this with setting up a `ref` and scrolling to the bottom - https://stackoverflow.com/a/42736127/9272042

Also if you are using `FlashList`, then you may use `renderScrollComponent` prop to specify your custom scroll component (i.e. just use example app code with all transformations/fake views).

psteinroe:
@kirillzyusko thanks for the hints. I boiled it down to `rotate: 180deg` not being applied to the children of the `ScrollView`. I now ditched all the transforms and use `scrollToBottom`. This is not optimal as the scrolling process is visible to the user, but it works and is good enough for now. Would be great to find a solution for this, although I understand this is is unrelated to this project.

```tsx
 const scrollViewStyle = useAnimatedStyle(
    () => ({
      height: '100%',
      backgroundColor: '#e2e8f0',
      transform: [{ translateY: height.value }, ...styles.inverted.transform],
    }),
    []
  );

[...]

PR #236, opened 2023-09-12
chore: add reanimated chat flatlist example

dayze:
## 📜 Description

Add an example for a chat with reanimated and flatlist 

## 💡 Motivation and Context

There is only an example with a ScrollView, in a real chat app it's more common to encounter a FlatList
It may solve this issue https://github.com/kirillzyusko/react-native-keyboard-controller/issues/156

Tested iPhone 14 Pro 

https://github.com/kirillzyusko/react-native-keyboard-controller/assets/11976270/9d73f4f0-e1a4-4ea6-a0f8-a870284c0322
## 📝 Checklist

- [ ] CI successfully passed


kirillzyusko:
Hi @dayze 🙌 

Thank you for the contribution! 😎 I checked and it seems like messages go in a wrong order. "Hmmm" message should be in top. The difference:

|ScrollView|FlatList|
|----------|--------|
|image|image|

Could you please have a look and fix it?
In a meantime I will try to test your changes on Android 👀

example/src/screens/Examples/ReanimatedChatFlatlist/index.tsx
```diff
@@ -0,0 +1,46 @@
+import React from 'react';
+import { FlatList, ListRenderItem, TextInput, View } from 'react-native';
+import { useReanimatedKeyboardAnimation } from 'react-native-keyboard-controller';
+import { useAnimatedStyle } from 'react-native-reanimated';
+
+import Message from '../../../components/Message';
+import { history } from '../../../components/Message/data';
+
+import type { StackScreenProps } from '@react-navigation/stack';
+import Animated from 'react-native-reanimated';
```
kirillzyusko, line 9: You already have `reanimated` import above - can you combine them in one line?

dayze:
Hello @kirillzyusko ! 
I've fixed the order of messages with a reverse on the array, it's necessary with the inverted FlatList
image

example/src/screens/Examples/ReanimatedChatFlatlist/index.tsx
```diff
@@ -0,0 +1,45 @@
...
+  const fakeView = useAnimatedStyle(
+    () => ({
+      height: Math.abs(height.value),
+    }),
+    []
+  );
+
+  const renderItem: ListRenderItem<MessageProps> = ({ item, index }) => {
+    return <Message key={index} {...item} />;
+  };
```
kirillzyusko, line 28: Can we make it a separate function outside of the render? Don't want to re-create this function on every re-render (thou I understand that in this example it'll not be recreated ever, because we are not using state or external props, but a lot of people may try to copy the code, so it would be better to create a proper design in advance 🙂).

example/src/screens/Examples/ReanimatedChatFlatlist/index.tsx
```diff
@@ -0,0 +1,45 @@
...
+
+import type { StackScreenProps } from '@react-navigation/stack';
+import Animated from 'react-native-reanimated';
+import { MessageProps } from '../../../components/Message/types';
+import type { ExamplesStackParamList } from '../../../navigation/ExamplesStack';
+import styles from './styles';
+
+type Props = StackScreenProps<ExamplesStackParamList>;
+
+function ReanimatedChatFlatlist({ navigation }: Props) {
```
kirillzyusko, line 16: I think navigation prop is unused here? Does it makes sense to remove it then?

kirillzyusko:
@dayze This is not related to this PR, but I want to ask from curiosity. I've noticed, that after RN 0.72 chat apps became laggy. Below is a comparison between different approaches for styling:

|Change height|translateY|
|--------------|-----------|
|<video src="https://github.com/kirillzyusko/react-native-keyboard-controller/assets/22820318/82487460-3bd7-47ea-aa01-fbc075c35d84">|<video src="https://github.com/kirillzyusko/react-native-keyboard-controller/assets/22820318/490c9b78-7e57-48af-92ed-1ff1d79378cd">|

<details>
  <summary>Approach for styling with <code>translateY</code></summary>
  
  ```tsx
  import React from 'react';
import {
  FlatList,
  ListRenderItem,
  Platform,
  TextInput,
  View,
} from 'react-native';
import {
  useKeyboardHandler,
  useReanimatedKeyboardAnimation,
} from 'react-native-keyboard-controller';
import {
  useAnimatedStyle,
  useSharedValue,
  withTiming,
  Easing,
} from 'react-native-reanimated';

import Message from '../../../components/Message';
import { history } from '../../../components/Message/data';

import type { StackScreenProps } from '@react-navigation/stack';
import Animated from 'react-native-reanimated';
import { MessageProps } from '../../../components/Message/types';
import type { ExamplesStackParamList } from '../../../navigation/ExamplesStack';
import styles from './styles';

type Props = StackScreenProps<ExamplesStackParamList>;

export const useTelegramTransitions = () => {
  const height = useSharedValue(0);

  useKeyboardHandler(
    {
      onStart: (e) => {
        'worklet';

        if (Platform.OS === 'android') {
          // on Android Telegram is not using androidx.core values and uses custom interpolation
          // duration is taken from here: https://github.com/DrKLO/Telegram/blob/e9a35cea54c06277c69d41b8e25d94b5d7ede065/TMessagesProj/src/main/java/org/telegram/ui/ActionBar/AdjustPanLayoutHelper.java#L39
          // and bezier is taken from: https://github.com/DrKLO/Telegram/blob/e9a35cea54c06277c69d41b8e25d94b5d7ede065/TMessagesProj/src/main/java/androidx/recyclerview/widget/ChatListItemAnimator.java#L40
          height.value = withTiming(-e.height, {
            duration: 250,
            easing: Easing.bezier(
              0.19919472913616398,
              0.010644531250000006,
              0.27920937042459737,
              0.91025390625
            ),
          });
        } else {
          // on iOS Telegram simply moves TextInput synchronously with the content
          // to achieve such behavior we are instantly change `height.value` to keyboard
          // final frame - iOS will schedule layout animation and it will move the content
          // altogether with the keyboard
          height.value = -e.height;
        }
      },
    },
    []
  );

  return { height };
};

function ReanimatedChatFlatlist({ navigation }: Props) {
  const { height } = useTelegramTransitions();

  const fakeView = useAnimatedStyle(
    () => ({
      transform: [{ translateY: height.value }],
    }),
    []
  );

  const renderItem: ListRenderItem<MessageProps> = ({ item, index }) => {
    return <Message key={index} {...item} />;
  };

  return (
    <Animated.View style={[styles.container, fakeView]}>
      <FlatList
        inverted
        initialNumToRender={15}
        contentContainerStyle={styles.contentContainer}
        data={history}
        renderItem={renderItem}
      />
      <TextInput style={styles.textInput} />
    </Animated.View>
  );
}

export default ReanimatedChatFlatlist;

  ```
</details>

Basically if we just do `translate` - then animation looks smoother. I understand, that animation `height` is more heavy operation than animation `translateY`, but still curious why the difference is so noticeable 🤔 Maybe you know anything about it?

kirillzyusko:
@dayze Please, fix CI and we're ready to go!

dayze:
> @dayze This is not related to this PR, but I want to ask from curiosity. I've noticed, that after RN 0.72 chat apps became laggy. Below is a comparison between different approaches for styling:
> 
> Change height	translateY
>  telegram-cloud-document-2-5190505081438481568.mp4 
>  telegram-cloud-document-2-5190505081438481570.mp4 
> Approach for styling with `translateY`
> Basically if we just do `translate` - then animation looks smoother. I understand, that animation `height` is more heavy operation than animation `translateY`, but still curious why the difference is so noticeable 🤔 Maybe you know anything about it?

I did notice it in our app too, but no idea for now. 
Like you I suspect the height computation to be more heavy 🤔

kirillzyusko:
@dayze I did a minor code modifications (didn't want to disturb you anymore). Thank you for the contribution! It was very helpful! ❤️

Dajust:
@dayze Thanks for this example, I struggled with this problem for days and couldn't get it without introducing more problems 😅

@kirillzyusko The problem with doing  `translate` is when you have only a few messages on the list. The messages disappear from view as the container moves up. Plus when the keyboard is open, you can't scroll up to the first message at the top.

On the other hand, animating the fakeView height didn't even work for me. The fakeView jumps out even before the keyboard appears. This stuff is hard.

Dajust:
To be clear about the height animation. When animating from 0 to `keyboard height` on iOS, the height _does not_ animate. But when closing the keyboard - animating from `keyboard height` to 0 - the `height` animation works just fine.

My workaround was to modify the `useTelegramTransition` hook to manually calculate the height transition for both platforms.

```js
export const useTelegramTransitions = () => {
  const height = useSharedValue(0);

  useKeyboardHandler(
    {
      onStart: e => {
        'worklet';
        
        
        // I moved this up here to work for both platforms.
        height.value = withTiming(-e.height, {
          duration: 250,
          easing: Easing.bezier(
            0.19919472913616398,
            0.010644531250000006,
            0.27920937042459737,
            0.91025390625,
          ),
        });

        if (Platform.OS === 'android') {
          // on Android Telegram is not using androidx.core values and uses custom interpolation
          // duration is taken from here: https://github.com/DrKLO/Telegram/blob/e9a35cea54c06277c69d41b8e25d94b5d7ede065/TMessagesProj/src/main/java/org/telegram/ui/ActionBar/AdjustPanLayoutHelper.java#L39
          // and bezier is taken from: https://github.com/DrKLO/Telegram/blob/e9a35cea54c06277c69d41b8e25d94b5d7ede065/TMessagesProj/src/main/java/androidx/recyclerview/widget/ChatListItemAnimator.java#L40
        } else {
          // on iOS Telegram simply moves TextInput synchronously with the content
          // to achieve such behavior we are instantly change `height.value` to keyboard
          // final frame - iOS will schedule layout animation and it will move the content
          // altogether with the keyboard
          // FOR SOME REASON THIS DOESN'T WORK WHEN ANIMATING FROM 0 TO KEYBOARD HEIGHT
          // height.value = -e.height;
        }
      },
    },
    [],
  );

  return {height};
};
```

Then some hacks on my `fakeView` style.
```js
 const {keyboardShown} = useKeyboard();
  const insets = useSafeAreaInsets();
  const {height} = useReanimatedKeyboardAnimation();
  const {height: t_height} = useTelegramTransitions();
  const fakeView = useAnimatedStyle(
    () => ({
      height: keyboardShown
        ? height.value
          ? Math.abs(height.value + insets.bottom)
          : height.value
        : t_height.value
        ? Math.abs(t_height.value + insets.bottom)
        : t_height.value,
    }),
    [keyboardShown],
  );
  ```
  
  So if keyboard is shown, the closing animation should use `height` from `useReanimatedKeyboardAnimation`, but when keyboard is opening, I use the manual animation.
  
 It's quite ugly but it gets the animation working for both closing and opening on iOS.

kirillzyusko:
> On the other hand, animating the fakeView height didn't even work for me on iOS. The fakeView jumps out even before the keyboard appears.

@Dajust are you using fabric/new architecture?

In paper example app I'm using hooks provided by this library and it works good. But I also noticed that on Fabric some animations are missing.

And you are right - `useReanimatedKeyboardAnimation` doesn't have intermediate values as pointed out in [documentation](https://kirillzyusko.github.io/react-native-keyboard-controller/docs/recipes/platform-differences#ios). But you can use `useKeyboardHandler` + `onMove` callback. This callback has a gradual/intermediate values so your keyboard animation should work fine if you just consume values from `onMove` callback 🙂

PR #252, opened 2023-10-04
refactor: integrate `typescript-eslint-plugin`

IvanIhnatsiuk:
## 📜 Description

added @typescript-eslint/eslint-plugin to support importing and exporting types with eslint to avoid adding some types file in the final build. Also, it would be useful to have a more strict eslint in the project 🙂 . 

## 💡 Motivation and Context

code base improvement
## 📢 Changelog

- Created .eslintrc.cjs with eslint configuration.
- Added @typescript-eslint/eslint-plugin, @typescript-eslint/parser and eslint-plugin-jest

Tested with the `yarn lint` command 🙂 . 

## 📝 Checklist

- [x] CI successfully passed

### DIFF
--- a/src/context.ts
+++ b/src/context.ts
@@ -1,9 +1,9 @@
-import type React from 'react';
 import { createContext, useContext } from 'react';
 import { Animated } from 'react-native';
 
-import type { SharedValue } from 'react-native-reanimated';
 import type { KeyboardHandlers } from './types';
+import type React from 'react';
+import type { SharedValue } from 'react-native-reanimated';
 
 export type AnimatedContext = {
   progress: Animated.Value;
--- a/src/context.ts
+++ b/src/context.ts
@@ -1,4 +1,5 @@
-import React, { createContext, useContext } from 'react';
+import type React from 'react';
+import { createContext, useContext } from 'react';
 import { Animated } from 'react-native';
 
 import type { SharedValue } from 'react-native-reanimated';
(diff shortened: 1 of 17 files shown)


src/specs/NativeStatusBarManagerCompat.ts
```diff
@@ -2,7 +2,7 @@ import type { TurboModule } from 'react-native';
 import { TurboModuleRegistry } from 'react-native';
 
 export interface Spec extends TurboModule {
-  readonly getConstants: () => {};
+  readonly getConstants: () => object;
```
kirillzyusko, line 5: @IvanIhnatsiuk can you revert to previous variant? ()=> object is not recognized by codegen and it'll break fabric builds.

src/context.ts
```diff
@@ -1,4 +1,5 @@
-import React, { createContext, useContext } from 'react';
+import type React from 'react';
```
kirillzyusko, line 1: Is it possible to have `type` import in the end of/after all imports?

IvanIhnatsiuk: This can be done using another eslint plugin. I can do it in this PR, but I think it would be better to create a separate PR for this 🙂

src/specs/NativeStatusBarManagerCompat.ts
IvanIhnatsiuk: 👀

src/context.ts
kirillzyusko: @IvanIhnatsiuk let's do in this, since anyway we are changing import style :)

PR #296, opened 2023-12-14
feat: added `KeyboardAwareScrollView`

kirillzyusko:
## 📜 Description

Added `KeyboardAwareScrollView` as a component available from a library.

## 💡 Motivation and Context

I got asked many times times to export this component from the package (now users had to copy the source code of this component into theirs apps).

Even though this component is not perfect and contains some bugs in edge cases I still think it's better to open-source it and fix reported bugs later (rather than keep it just in example app and people will not be aware about its existence).

### JS
- exported `KeyboardAwareScrollView` from the package;
- removed code from examples app and start to use a component provided from the package;

### Docs
- added a page about new component;

e2e tests, page preview.

## 📝 Checklist

- [x] CI successfully passed


src/components/KeyboardAwareScrollView/useSmoothKeyboardHandler.ts
```diff
@@ -5,6 +5,7 @@ import {
   useSharedValue,
   withTiming,
 } from 'react-native-reanimated';
+
```
kirillzyusko, line 8: @IvanIhnatsiuk any ideas why this empty line is required? 👀

PR #308, opened 2023-12-23
fix: use `install_modules_dependencies` to fix compilation issues on RN 0.73 with Fabric

kirillzyusko:
## 📜 Description

Started to use `install_modules_dependencies` in `.podspec`.

## 💡 Motivation and Context

It's vital to use `install_modules_dependencies` because it simplifies integration process (this function is shipped in RN, so RN team cares about internal deps that could be provided to the project).

So it means that we don't need to specify it manually anymore and instead we can rely on the implementation provided by the package.

However we need to keep in mind, that this function is available only from RN 0.71, so if we support RN < 0.71 we need to have fallback code.

The other aspect that was discovered during the implementation is the fact that this function was broken on RN 0.72 + Swift + New Arch combination, but with latest RN release it's not broken anymore (that's the reason why I updated a RN dependency in this PR - to verify that it actually works as expected).

Closes https://github.com/kirillzyusko/react-native-keyboard-controller/issues/371

### JS
- updated RN to 0.72.11

### iOS
- modified `.podspec` file to use `install_modules_dependencies` function.

Tested manually in reproduction example + via CI.

## 📝 Checklist

- [x] CI successfully passed

### DIFF
--- a/FabricExample/Gemfile
+++ b/FabricExample/Gemfile
@@ -3,4 +3,5 @@ source 'https://rubygems.org'
 # You may use http://rbenv.org/ or https://rvm.io/ to install and use this version
 ruby ">= 2.6.10"
 
-gem 'cocoapods', '~> 1.12'
+gem 'cocoapods', '~> 1.13'
+gem 'activesupport', '>= 6.1.7.3', '< 7.1.0'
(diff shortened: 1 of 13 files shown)


kirillzyusko:
Seems like fabric build is failing for RN < 0.73. I asked a question here https://github.com/lottie-react-native/lottie-react-native/pull/1139#issuecomment-1868550486 👀

kirillzyusko:
We need to wait for new RN 0.72 version to be released and after that we can update RN version in RNKC and merge this PR.

PR #313, opened 2023-12-25
Proposal: Apply prettier on the whole project files

zoontek:
## 📜 Description

This is a proposal. When I'm working on the repository, I notice that the formatter (prettier) update some part of the files even if I don't changed them. This occurs because some files are not formatted with it, this PR add a new npm command (`format`) and run it.

I know that this is a lot of changes (especially in the docs markdown files), feels free to close the PR if you don't want to deal with this.

## 💡 Motivation and Context

This simplifies collaboration on the project, can even been run on commit using `lint-staged`.


FabricExample/__tests__/focused-input-handler.spec.tsx
```diff
@@ -17,7 +22,7 @@ function WhatUserTyped() {
     },
   });
 
-  return <Text testID='text'>{text}</Text>;
+  return <Text testID="text">{text}</Text>;
```
kirillzyusko, line 25: Can't we force double quotes everywhere?

In this PR import-statements have single quote, jsx string props have double quotes, plain string have single quotes.

I'd love to see the consistency here and use double quotes if possible (it may have a conflicts with existing eslint/linter rules, but feel free to change it)

P. S. I love double quotes because in java single and double quotes have different meaning (symbol vs string) so I like to have a consistency between two languages

docs/docs/api/hooks/input/use-reanimated-focused-input.md
```diff
@@ -23,22 +31,24 @@ The `input` property from this hook is returned as `SharedValue`. The returned d
 
 ```ts
 type FocusedInputLayoutChangedEvent = {
-  target: number; // tag of the focused TextInput
-  layout: { // layout of the focused TextInput
+  // tag of the focused TextInput
```
kirillzyusko, line 34: Slightly strange, that the comment was moved above, but comments below were not 😐

docs/docs/api/keyboard-provider.md
```diff
@@ -30,13 +30,11 @@ Could be useful to set it to `false` if you want to activate the module only on
...
 ```tsx
-import { KeyboardProvider } from "react-native-keyboard-controller";
+import { KeyboardProvider } from 'react-native-keyboard-controller';
 
 const App = () => {
   return (
-    <KeyboardProvider>
-      {/* other components in your tree */}
-    </KeyboardProvider>
+    <KeyboardProvider>{/* other components in your tree */}</KeyboardProvider>
```
kirillzyusko, line 37: image

Looks slightly truncated 😔

kirillzyusko:
Thank you @zoontek for your PR
I knew that only code in src was properly formatted, but somehow ignored it 🙈 

Overall I like the idea of auto-formatting and definitely would like to merge this PR. But I have some questions:
- could you rebase to latest changes and apply prettier again? I just prepared a new release so some new files were added recently.
- could you add GH action that will verify that all code is properly formatted? (I like the usage of git-hooks, but some external contributors may push/commit with `--no-verify` flag, so I'd like to have an independent verifier in CI)

FabricExample/__tests__/focused-input-handler.spec.tsx
zoontek: Sure, the current prettier config is `"singleQuote": true`, let's switch to the default (`false`)

docs/docs/api/hooks/input/use-reanimated-focused-input.md
zoontek: I'll update to prettier v3 to see if we can keep this behavior (comment on the same line, after an open bracket)

docs/docs/api/keyboard-provider.md
zoontek: I added `The ` to enforce new lines 😄

docs/docs/api/hooks/input/use-reanimated-focused-input.md
zoontek: I cannot keep it on the same line, so I kept it this way:

```ts
type FocusedInputLayoutChangedEvent = {
  target: number; // tag of the focused TextInput

  // layout of the focused TextInput
  layout: {
    x: number; // `x` coordinate inside the parent component
    y: number; // `y` coordinate inside the parent component
    width: number; // `width` of the TextInput
    height: number; // `height` of the TextInput
    absoluteX: number; // `x` coordinate on the screen
    absoluteY: number; // `y` coordinate on the screen
  };
};
```

Are you OK with this?

kirillzyusko: @zoontek yes, I'm okay with that 🙂

zoontek:
> could you add GH action that will verify that all code is properly formatted? (I like the usage of git-hooks, but some external contributors may push/commit with --no-verify flag, so I'd like to have an independent verifier in CI)

There's already `eslint-plugin-prettier` on the project, maybe we should extend the linting to the whole project? (by removing `docs`, `example` and `FabricExample` from `ignorePatterns`)

Screenshot 2023-12-26 at 20 43 46

There's currently a lot of issues with the `import/order` rule if you do this, but we can run lint with `--fix` to clean that.

I did it in this commit: https://github.com/kirillzyusko/react-native-keyboard-controller/pull/313/commits/1d6f159c82fdbfe6dd6e006c923edf28db654c5f

This makes a lot of changes, but rebasing is really easy: never pick the changes from this branch in code, run `yarn format` and `yarn lint --fix` once the rebase is done.

zoontek:
- I updated `@react-native/eslint-config` to the latest version, aligned dependencies to match the config deps (`eslint-config-prettier`, `eslint-plugin-jest` and `eslint-plugin-prettier`)
- I had to bump actions to use node v18.x as v16.x is deprecated with this version of the eslint config
- I updated the prettier config to use `singleQuote: false`, `tabWidth: 2`, `useTabs: false` (the defaults in prettier v2) and `trailingComma: "all"` (the default in prettier v3)
- I fixed the few linting errors which were not `import/order` errors

kirillzyusko:
> maybe we should extend the linting to the whole project? (by removing docs, example and FabricExample from ignorePatterns)

@zoontek yes, that's definitely a good point (and I've seen you already did that 🎉 ). The reason why I want to have a prettier as CI job is because `eslint` is scanning only JS/TS files, but I see that prettier also fixes problems in `.md`/`.yml` and other files (which is great).

So I think having a prettier on CI would be good, do you agree?

I'm happy to merge all these changes (though I haven't reviewed latest changes yet), but want to have prettier job in place to assure that all new code will be properly formatted 🙂

zoontek:
@kirillzyusko I added a "Check formatting" step, but in the `verify-js.yml` workflow, even if it's not only for JS.

It's a bit of a duplicate with the `eslint` plugin, but it's allow formatting issues reporting in dev when the dev has a an editor with `eslint` configured.

kirillzyusko:
Awesome @zoontek 

I'll review everything again today evening (and maybe do some other fixes on my own, do not want to disturb you once again) and then I'll merge this PR 😎 

Thank you again for the contribution!

package.json
```diff
@@ -71,15 +72,15 @@
     "clang-format": "^1.8.0",
     "commitlint": "^11.0.0",
     "eslint": "^8.19.0",
-    "eslint-config-prettier": "^7.0.0",
+    "eslint-config-prettier": "^8.5.0",
     "eslint-import-resolver-typescript": "^3.6.1",
     "eslint-plugin-import": "^2.28.1",
-    "eslint-plugin-jest": "^27.4.2",
-    "eslint-plugin-prettier": "^3.1.3",
+    "eslint-plugin-jest": "^26.5.3",
```
kirillzyusko, line 78: @zoontek may I ask you why it was needed to downgrade `jest` plugin version?

zoontek: To align with [`@react-native/eslint-config` dependencies version](https://www.npmjs.com/package/@react-native/eslint-config/v/0.74.0?activeTab=code) and avoid installing two different versions (you can check in the `yarn.lock` that it performs a dedup)

kirillzyusko: @zoontek thanks, it makes sense 🙂 👍 (I haven't checked `yarn.lock` diff yet, but I'm in a progress 👀)

Issue #319, opened 2024-01-05
Jest mock doesn't include newly exported KeyboardAwareScrollView


bcgilliom:
**Describe the bug**

tests fail when `KeyboardAwareScrollView` is used in tests using the following from the docs

```
jest.mock("react-native-keyboard-controller", () =>
  require("react-native-keyboard-controller/jest"),
);
```

Adding something like:

```
export const KeyboardAwareScrollView = (props) => <ScrollView {...props} />;
```

to the mocks seems to work

**To Reproduce**

1. Add a test for a component that uses KASV
2. Run tests
3. see error about no being linked

**Expected behavior**
tests should run using the OOTB config

**Screenshots**
N/A

**Smartphone (please complete the following information):**

- Desktop OS: Mac
- Device: jest

kirillzyusko:
Thanks @bcgilliom 

I will have a look on it over weekend (actually I think I mocked all low level hooks/views, and `KeyboardAwareScrollView` just consumes them, so **theoretically** it should work, but I haven't tested it).

And yes, thank you for raising this issue 🙏

kirillzyusko:
@bcgilliom may I kindly ask you to check/review PR https://github.com/kirillzyusko/react-native-keyboard-controller/pull/320?

I've added `View` as a mock for `KeyboardAvoidingView` and `KeyboardStickyView` and added `ScrollView` as a mock for `KeyboardAwareScrollView`. But I don't know whether it's a good practice, because in snapshot testing non-view props (such as `behavior` for `KeyboardAvoidingView`) will be added to snapshot.

From one point it's good, because if someone accidentally changes it -> unit tests will fail, but I've tried to use default `KeyboardAvoidingView` and it's not including custom props into snapshot (Because a real implementation of `KeyboardAvoidingView` is used)

PR #320, opened 2024-01-06
fix: missing components mocks

kirillzyusko:
## 📜 Description

Added mocks for exported components, such as `KeyboardAvoidingView`, `KeyboardAwareScrollView`, `KeyboardStickyView`.

## 💡 Motivation and Context

The library should support mocks for exported components out of the box.

In an ideal world I'd like to mock dependencies that depends on native implementation and use a real implementation for components. However since we mock entire module (`react-native-keyboard-controller`) it becomes impossible to mock it partially. 

Even with `jest.requireActual()` for components import from `react-native-keyboard-controller` produced `{}` and as a result `useKeyboardHandler` and other native dependencies were undefined and it causes errors during component rendering.

Also a controversial topic was how to mock these components - as `View`/`ScrollView` from RN or as strings. I had at implementation of other popular repositories and came to conclusion, that it should be mocked as RN components (as a reference I used `react-native-gesture-handler`).

Closes https://github.com/kirillzyusko/react-native-keyboard-controller/issues/319

### Docs
- added new checkbox for PR template (to verify mocks);

### CI
- use latest version of sponsor labeling, because old one uses deprecated Node 12;

### JS

- added mocks for `KeyboardAvoidingView`, `KeyboardAwareScrollView`, `KeyboardStickyView`;
- use imports from `react-native-keyboard-controller` in `components`;

Added new snapshot tests to Fabric and paper examples.

## 📝 Checklist

- [x] CI successfully passed

### DIFF
--- a/jest/index.js
+++ b/jest/index.js
@@ -28,13 +28,19 @@ const focusedInput = {
 
 const mock = {
   // hooks
+  /// keyboard
   useKeyboardAnimation: jest.fn().mockReturnValue(values.animated),
   useReanimatedKeyboardAnimation: jest.fn().mockReturnValue(values.reanimated),
   useResizeMode: jest.fn(),
   useGenericKeyboardHandler: jest.fn(),
   useKeyboardHandler: jest.fn(),
+  /// input
   useReanimatedFocusedInput: jest.fn().mockReturnValue(focusedInput),
   useFocusedInputHandler: jest.fn(),
+  /// module
+  useKeyboardController: jest
+    .fn()
+    .mockReturnValue({ setEnabled: jest.fn(), enabled: true }),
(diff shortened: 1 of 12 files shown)


jest/index.js
```diff
@@ -49,6 +49,10 @@ const mock = {
   KeyboardGestureArea: "KeyboardGestureArea",
   // providers
   KeyboardProvider: "KeyboardProvider",
+  // components
+  KeyboardStickyView: View,
+  KeyboardAvoidingView: View,
+  KeyboardAwareScrollView: ScrollView,
```
kirillzyusko, line 55: Should we add mock for `useKeyboardController` hook as well?

PR #321, opened 2024-01-08
fix: remove wrapper view from `KeyboardAwareScrollView`

kirillzyusko:
## 📜 Description

Removed view wrapper. Attached that view as last child + kept padding for this view.

## 💡 Motivation and Context

The approach with additional view was added in https://github.com/kirillzyusko/react-native-keyboard-controller/pull/257

I don't remember exact reason, but for me it seems like it was added, because TExtInputs were not able to grow. However such approach is causing additional issues:

- breaks styling in some cases
- stickyHeaderIndices={[0]} produces a crash (because RN will try to attach `Animated`-based style to REA view)

So in this PR I'm removing this view wrapper and apply additional padding in mostly the same way as it was before https://github.com/kirillzyusko/react-native-keyboard-controller/pull/257 (additional child-view in the end of ScrollView, but instead of `height` I'm animating `paddingBottom`).

I've tested and it still works (i. e. inputs are growing). E2E tests also passing (so there seems to be 1px difference, which can be neglected, but i had to update assets to assure E2E tests consistency).

> [!IMPORTANT]
> This fix will introduce https://github.com/software-mansion/react-native-reanimated/issues/5567. I've added that issue as "known issue" in docs page. 
I think it's better to have properly working paper architecture and Fabric can be fixed later (because it's still in early adoption and is not widely used).

Closes https://github.com/kirillzyusko/react-native-keyboard-controller/issues/325

### E2E

- added `detox-clean` command (useful when you updated XCode version and haven't run detox tests yet)

### JS

- removed wrapper and moved view as last child

Tested on:
- e2e (Android, iOS - paper)
- Pixel 7 Pro (Android 14, paper)

There is no visual difference 🙂 

## 📝 Checklist

- [x] CI successfully passed


kirillzyusko:
For history purposes I'm attaching pixel diff (it seems like it under/over-scrolls 1px, what can be ignored):

AwareScrollViewKeyboardClosed-diff

kirillzyusko:
I opened an issue for Fabric problem in REA repository: https://github.com/software-mansion/react-native-reanimated/issues/5567

PR #328, opened 2024-01-12
feat: add setting to KASV to disable resetting scroll when keyboard closes

bcgilliom:
## 📜 Description

Add a prop to allow disabling resetting the scroll position (similar I think to KASV `resetScrollToCoords`)

## 💡 Motivation and Context

In our app, it was causing some issues to have the scroll reset when the keyboard closes. For example if the user focuses an input, and the while the keyboard is open, taps an input that opens a bottom sheet picker, the field they just tapped on can get scrolled off screen

One slight issue in the example (doesn't seem to happen in our app...) is that when the input is at the very bottom, the scrollview kind of jumps when the keyboard finishes closing (can be seen at end of video)

just let me know if there's a better way to handle this and I can try to update the PR. I didn't add a case for the new prop to unit tests because it didn't look like `bottomOffset` had one

### JS

- add a new prop to KASV `resetScroll` - defaults to true to maintain current behavior

- tested in example app

updated video 1/16:

https://github.com/kirillzyusko/react-native-keyboard-controller/assets/5893743/68d08876-6253-4a49-9166-73f9d4d5ba46

## 📝 Checklist

- [x] CI successfully passed


src/components/KeyboardAwareScrollView/index.tsx
```diff
@@ -67,18 +68,25 @@ type KeyboardAwareScrollViewProps = {
...
+  const keyboardWillAppear = useSharedValue(false);
   const tag = useSharedValue(-1);
   const initialKeyboardSize = useSharedValue(0);
   const scrollBeforeKeyboardMovement = useSharedValue(0);
+  const resetScrollPosition = useSharedValue(resetScroll);
   const { input } = useReanimatedFocusedInput();
   const layout = useSharedValue<FocusedInputLayoutChangedEvent | null>(null);
 
+  useEffect(() => {
+    resetScrollPosition.value = resetScroll;
```
bcgilliom, line 87: the prop changes don't get picked up by the worklets

src/components/KeyboardAwareScrollView/index.tsx
```diff
@@ -24,6 +24,7 @@ import type { FocusedInputLayoutChangedEvent } from "react-native-keyboard-contr
 
 type KeyboardAwareScrollViewProps = {
   bottomOffset?: number;
+  resetScroll?: boolean;
```
bcgilliom, line 27: maybe could be `disableScrollreset = false`?

kirillzyusko:
Hello @bcgilliom 

Thank you for your PR!

> One slight issue in the example (doesn't seem to happen in our app...) is that when the input is at the very bottom, the scrollview kind of jumps when the keyboard finishes closing (can be seen at end of video)

I fixed it in https://github.com/kirillzyusko/react-native-keyboard-controller/pull/332 and because of that this PR has conflicts 😓 Would you mind to resolve them?

> just let me know if there's a better way to handle this and I can try to update the PR.

I'm just thinking, that maybe it would be easier/more convenient to create a generic-purpose solution and add `enabled` property? 🤔 If it's `enabled={false}` (we can just add a check in `maybeScroll` function) then it shouldn't do any scrolling at all? And you can disable KASV when you need (i. e. you are going to show bottom-sheet) - what do you think about it?

bcgilliom:
> I fixed it in #332 and because of that this PR has conflicts 😓 Would you mind to resolve them?

- yep, will do

> I'm just thinking, that maybe it would be easier/more convenient to create a generic-purpose solution and add `enabled` property? 🤔 If it's `enabled={false}` (we can just add a check in `maybeScroll` function) then it shouldn't do any scrolling at all? And you can disable KASV when you need (i. e. you are going to show bottom-sheet) - what do you think about it?

For our use case, we still need it to work when the keyboard opens, we just generally don't want the behavior that scrolls the input back down to it's original position when the keyboard closes. (as far as I can tell, that's the behavior Apple uses in something like calendar - you can try editing an event a tapping the URL field near the bottom)

src/components/KeyboardAwareScrollView/index.tsx
kirillzyusko: @bcgilliom Even if we add this property to dependency array of worklet hooks, i. e. to `useKeyboardHandler`?

bcgilliom: ohh, I might have missed that - will try 👍

bcgilliom: ok, that's weird - I swear I tried that, but seems to be working - update incoming :-)

bcgilliom:
@kirillzyusko updated - just let me know if I missed anything or I can do more to help!

kirillzyusko:
Hi @bcgilliom 

Overall it looks good, I'll try to find some time for writing e2e tests and documentation - I hope to merge this PR this week (but maybe will do it only on weekend, because right now have a lot of work).

And yes, thank you for your contribution! 😊

kirillzyusko:
Merged! Thanks @bcgilliom for your contribution 🙌

Issue #303, opened 2023-12-20
KeyboardAccesoryView Input navigation arrows  


VladyslavMartynov10:
In some recent applications, I've noticed KeyboardAccessoryView with navigation prev/next arrows & action done button. I understand that this was designed for enhanced user accessibility and quicker navigation between input fields, eliminating the need to scroll down to focus on the necessary input.

I was searching for references that could potentially be used for implementation and found only this library:
https://github.com/douglasjunior/react-native-keyboard-manager. The main issue with this implementation is that it's available only on iOS, and in some cases, it cannot be fully controlled by the developer.

Moreover, after a couple of hours, I created a KeyboardAccessoryView based on `KeyboardStickyView` right out of the box, resulting in the following component:

```tsx
import React, { useMemo } from 'react';
import { Keyboard, StyleSheet, TouchableOpacity, View, Text } from 'react-native';
import {
  KeyboardStickyView,
  useReanimatedKeyboardAnimation,
} from 'react-native-keyboard-controller';
import Animated, {
  Extrapolate,
  interpolate,
  useAnimatedStyle,
} from 'react-native-reanimated';
import { Icon } from 'react-native-vector-icons';


interface KeyboardAccessoryViewProps {
  onNextPress?: () => void;
  onPrevPress?: () => void;
  onDoneButtonPress?: () => void;
}

const styles = StyleSheet.create({
  container: {
    flexDirection: 'row'
  }
});

export const KeyboardAccessoryView: React.FC<KeyboardAccessoryViewProps> = ({
  onNextPress,
  onPrevPress,
  onDoneButtonPress,
}) => {
  const offset = useMemo(() => ({ closed: 0, opened: 0 }), []);

  const { progress } = useReanimatedKeyboardAnimation();

  const onDonePress = () => {
    Keyboard.dismiss();
  };

  const onNext = () => {
    onNextPress?.();
  };

  const onPrev = () => {
    onPrevPress?.();
  };

  const animatedStyle = useAnimatedStyle(() => {
    return {
      opacity: interpolate(progress.value, [0, 1], [0, 1], Extrapolate.CLAMP),
      transform: [
        {
          translateY: interpolate(
            progress.value,
            [0, 1],
            [30, 0],
            Extrapolate.CLAMP,
          ),
        },
      ],
    };
  });

  return (
    <Animated.View style={[animatedStyle]}>
      <KeyboardStickyView offset={offset} style={styles.container}>

[...]

PR #346, opened 2024-01-29
feat: keyboard toolbar

kirillzyusko:
## 📜 Description

Added keyboard toolbar component.

## 💡 Motivation and Context

The feature was requested in https://github.com/kirillzyusko/react-native-keyboard-controller/issues/303

Before the implementation I decided to make a research to see how similar tasks are implemented on other platforms/framework. I went through:
- https://pub.dev/packages/keyboard_actions
- https://github.com/hackiftekhar/IQKeyboardManager
- default toolbar on iOS (on web)

The downside of `keyboard_actions` from Flutter is the fact that it requires a lot of hand engineering (setting `refs` to each input and the passing this array to the keyboard). And we need to do that for each screen.

However `IQKeyboardManager` handles it automatically via view traversal. I thought that a similar algorithm can be implemented in a react-native, because all JS views are shown in a native view hierarchy as corresponding native views.

The `IQKeyboardManager` offers a lot of customization which is not needed in `react-native` ecosystem (like specifying array of `UIVIewController`s where `KeyboardToolbar` should be hidden - it should be done via conditional rendering in RN world). On the other side - solution from `IQKeyboardManager` is not very customizable in `react-native` ecosystem (we can not specify rendering for custom elements).

So I took inspiration from all frameworks and decided to create a new approach that will ideally fit `react-native` projects - it'll not require a lot of hand engineering and will offer a great flexibility and customization. The ideal goal is to have the API as on web (from simplicity perspective) and to be a very cusomizable.

`IQKeyboardManager` handles a lot of cases, but I decided to start from something basic and then enhance the API. So the core API is a new method for `KeyboardController` - `setFocusTo("next" | "prev")`. It'll go to the next/prev field from currently focused input. For that I implemented code that goes through view hierarchy on both platforms.

Later I started to analyze further specifications and decided to add `setFocusTo("current")` API that will bring a keyboard back (set focus to last focused input).

Based on this API I created a precise replica of `KeyboardToolbar` (took a reference from `web` on iOS) and tried to create something that looks similar on Android (overall matches iOS design but uses colors that matches system UI on Android).

Remaining thing was how to tell JS when next/prev buttons available/active. I had few options on which way to use for dispatching:
- as plain events via `eventEmitter` (which I selected in the end);
- via `KeyboardControllerView` (we would need to add data in context, and it's kind of strange to have a lot of unrelated data/methods in context, we start to break SOLID principles here);
- via new `ToolbarView` (it could help to propagate events synchronously via reanimated-event-bus, but we also have to use conditional rendering/specify other props, so I thought it would be more convenient to use plain state - this component should be highly customizable and setting props via `useAnimatedProps` may be not a trivial thing for devs who just came to RN world);

I decided to use first approach because it's very simple and it looks like it's good enough (in future it can be changed, I didn't make the API for events propagation public yet).

I also had a choice of which data structures to use:

```ts
type FocusDidSetEvent = {
  hasNext: boolean;
  hasPrev: boolean;
}
```

vs

```ts
type FocusDidSetEvent = {
  current: number;
  count: number;
}
```

I selected a second one, because in my understanding it may give a better flexibility (though for this approach we have to gather all inputs on a screen which is more resource consuming operation than finding prev and next `TextInput` - again now it works fine so let's keep it, but in future internal events shape can be changed (for example if we encounter performance problems) - these events not part of public API. In public API we just propagate `disabled` property - the way how it's calculated it's totally up to library).

Closes https://github.com/kirillzyusko/react-native-keyboard-controller/issues/303 https://github.com/facebook/react-native/issues/641

### Docs

- added a page for `KeyboardToolbar` API;
- added lottie animation for `KeyboardToolbar` usage;

### JS

- added `KeyboardToolbar` component;
- implemented `setFocusTo` method (specs, TS declarations);
- return an initialized variable from `interpolate` rather than an assigned shared value because it seems like setting happens asynchronously and initial render may be incorrect (if you add `console.log({ a: interpolation, b: lastInterpolation.value });` you will see next output `{"a": 42, "b": 0}, {"a": 42, "b": 0}, {"a": 42, "b": 42}`);

### iOS

- added `FocusedInputHolder` class (holds a reference to last focused input);
- added `ViewHierarchyNavigator` which travers view hierarchy back and forth;
- added new `setFocusTo` method to `KeyboardController`;

### Android

- use Java 17 and Gradle Plugin v7.2;
- added unit tests (using `robolectric`) for `ViewHierarchyNavigator` class;
- added `FocusedInputHolder` class (holds a reference to last focused input);
- added `ViewHierarchyNavigator` which travers view hierarchy back and forth;
- added new `setFocusTo` method to `KeyboardController`;

### E2E

- added new test for covering `KeyboardToolbar` functionality;
- added 2 new functions `expectElementBitmapsToBeEqual` and `doActionNTimes`;

### CI

- added a job for running Android unit tests

Tested manually on:
- Pixel 7 Pro (Android 14, real device);
- Xiaomi Redmi Note 5 Pro (Android 9, real device);
- Pixel 3A (Android 13, emulator);
- iPhone 15 (iOS 17.2, simulator)
- iPhone 15 Pro (iOS 17.2, simulator)

### Screenshots

|Android (light theme)|Android (dark theme)|iOS (light theme)|iOS (dark theme)|
|---------------------|----------------------|-----------------|-----------------|
|||||

### Videos

https://github.com/kirillzyusko/react-native-keyboard-controller/assets/22820318/dc05191f-1cf1-41e3-8068-f4316de40a95

## 📝 Checklist

- [x] CI successfully passed


kirillzyusko:
Basically it works, but the current codebase is far from being ideal (CI explicitly shows that 😁), but main algorithm for view traversal should be working 👀 

@VladyslavMartynov10 may I ask you to try to integrate it into your project and give a feedback of which features are missing/what could be improved?

@IvanIhnatsiuk tagging you since you also wanted to check an implementation and do some experiments (also hope for your help with TouchableNativeFeedback on Android)

src/components/KeyboardToolbar/index.tsx
```diff
@@ -0,0 +1,130 @@
...
+    });
+
+    return () => subscription.remove();
+  }, []);
+
+  return (
+    <KeyboardStickyView offset={{ closed: height }}>
+      <View
+        // TODO: check whether it's readable or not
+        accessibilityLabel="Toolbar"
```
IvanIhnatsiuk, line 58: I don't think that it will be informative to announce this accessibility label.

src/components/KeyboardToolbar/index.tsx
```diff
@@ -0,0 +1,130 @@
...
+        onLayout={onLayout}
```
IvanIhnatsiuk, line 59: if we have fixed size, can we remove layout calculation?

src/components/KeyboardToolbar/index.tsx
```diff
@@ -0,0 +1,130 @@
...
+        style={[
+          styles.toolbar,
+          height === 0 ? { marginBottom: -9999 } : null,
+          background,
+        ]}
+        testID="keyboard.toolbar"
```
IvanIhnatsiuk, line 65: could you move this testIdPrefix outside and re-use it? I'm asking it because it repeats several times

src/components/KeyboardToolbar/index.tsx
```diff
@@ -0,0 +1,130 @@
...
+          testID="keyboard.toolbar.next"
+        >
+          <Arrow disabled={isNextDisabled} direction="down" />
+        </Button>
+
+        <View style={styles.flex} testID="keyboard.toolbar.content">
+          {Content ? <Content /> : null}
+        </View>
+        <Button
+          accessibilityLabel="Done"
```
IvanIhnatsiuk, line 90: There is no any sense to add this a11y label because it's the same as text

src/components/KeyboardToolbar/index.tsx
```diff
@@ -0,0 +1,130 @@
...
+          testID="keyboard.toolbar.done"
+        >
+          <Text
+            style={[
+              {
+                marginRight: 8,
+                fontWeight: "600",
+                fontSize: 15,
+              },
+              done,
```
IvanIhnatsiuk, line 102: move static styles to Stylesheet and wrap styles with memo to avoid re-renders

src/components/KeyboardToolbar/index.tsx
```diff
@@ -0,0 +1,130 @@
...
+  return (
+    <KeyboardStickyView offset={{ closed: height }}>
+      <View
+        // TODO: check whether it's readable or not
+        accessibilityLabel="Toolbar"
+        onLayout={onLayout}
+        style={[
+          styles.toolbar,
+          height === 0 ? { marginBottom: -9999 } : null,
+          background,
```
IvanIhnatsiuk, line 63: could you explain why do we need to set marginBottom if height is 0? Also, I think that we don't need height and it can be removed in the future

src/components/KeyboardToolbar/index.tsx
```diff
@@ -0,0 +1,130 @@
...
+  const onLayout = useCallback((e: LayoutChangeEvent) => {
+    setHeight(e.nativeEvent.layout.height);
+  }, []);
+
+  useEffect(() => {
+    const subscription = FocusedInputEvents.addListener("focusDidSet", (e) => {
+      setInputs(e);
+    });
+
+    return () => subscription.remove();
```
IvanIhnatsiuk, line 51:
```suggestion
    return subscription.remove;
```

src/components/KeyboardToolbar/colors.ts
```diff
@@ -0,0 +1,34 @@
+import { Platform, PlatformColor } from "react-native";
+
+type Color = ReturnType<typeof PlatformColor> | string;
```
IvanIhnatsiuk, line 3: you can use ColorValue type from react-native

VladyslavMartynov10:
Hello everyone! In my opinion, this first version of the KeyboardToolBar is functioning quite well and covers most of the basic scenarios.

Here's a list of potential improvements:

- Add support for dynamic placeholder text in the middle of the toolbar.
- Customize navigation arrows and the 'done' button (the React component part).
- Conduct testing with a large number of inputs, for example, 100 or more.

I've made a brief video demonstrating how it operates on a Samsung device with the One Ui system. Tested both in 60/120 FPS mode. Here are results:

60 FPS:

https://github.com/kirillzyusko/react-native-keyboard-controller/assets/115457344/d4648f25-5f21-445a-b6fb-74d5fe977525

120FPS:

https://github.com/kirillzyusko/react-native-keyboard-controller/assets/115457344/51f889d8-4788-4329-ac5a-e9a664668418

kirillzyusko:
Hello @VladyslavMartynov10  👋 

> Add support for dynamic placeholder text in the middle of the toolbar.

I think it's already doable via `Content` property? `AutoFill Contacts` is implemented using this prop.

> Customize navigation arrows and the 'done' button (the React component part).

Agree, I think we need to:
- give an ability to render custom touchables;
- give an ability to render custom content;
- give an ability to modify colors/text

> Conduct testing with a large number of inputs, for example, 100 or more.

I think the amount of input doesn't matter, though from performance perspective it would be definetely a good point!

> Support landscape mode

Need to check that!

Regarding FPS count - I think it doesn't change a lot between 60 and 120 FPS. Am I right?

VladyslavMartynov10:
Hey @kirillzyusko !

In comparison to my initial JS version which takes about average 7-12 PFS lost, native implementation definitely takes the lead 😁.

kirillzyusko:
> In comparison to my initial JS version which takes about average 7-12 PFS lost, native implementation definitely takes the lead 😁.

Ah, glad to know, that native implementation works faster 😎 

I'll take into consideration all feedbacks that you provided and when new version will be avaialble for testing I'll ask you once again to test it 😊

e2e/kit/004-keyboard-toolbar.e2e.ts
```diff
@@ -0,0 +1,82 @@
...
+    );
+  });
+
+  it("should skip a disabled fields", async () => {
+    await waitAndTap("keyboard.toolbar.next");
+    await expect(element(by.id("TextInput#5"))).toBeFocused();
+  });
+
+  it("should handle multiple clicks in row", async () => {
+    await doActionNTimes(() => waitAndTap("keyboard.toolbar.next"), 3);
```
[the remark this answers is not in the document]
kirillzyusko: @IvanIhnatsiuk `multipleTap` taps very fast and only one tap actually gets recognized by `Touchable`, so I'm reverting back to `doActionNTimes` usage

src/components/KeyboardToolbar/index.tsx
kirillzyusko: @IvanIhnatsiuk I know, but i just want to have a better code consistency. Let's leave it for now 😉

kirillzyusko:
@VladyslavMartynov10 @IvanIhnatsiuk I fixed all the comments (and overall current version looks much more stable than a previous version). Would you mind to test it and give your feedback? 👀

docs/docs/api/components/keyboard-toolbar/index.mdx
```diff
@@ -0,0 +1,134 @@
...
+
+:::info dark/light theme support
+Don't forget that you need to specify colors for **both** `dark` and `light` theme. The theme will be selected automatically based on the device preferences.
+:::
+
+## Example
+
+```tsx
+
+```
```
kirillzyusko, line 129: Add example

src/bindings.native.ts
```diff
@@ -28,11 +29,16 @@ export const KeyboardController = (
...
 
+const KEYBOARD_CONTROLLER_NAMESPACE = "KeyboardController::";
 const eventEmitter = new NativeEventEmitter(KeyboardController);
 
 export const KeyboardEvents: KeyboardEventsModule = {
   addListener: (name, cb) =>
-    eventEmitter.addListener("KeyboardController::" + name, cb),
+    eventEmitter.addListener(KEYBOARD_CONTROLLER_NAMESPACE + name, cb),
+};
+export const FocusedInputEvents: FocusedInputEventsModule = {
```
kirillzyusko, line 39: Let's add a JSDoc saying this is for internal usage only and is a subject to breaking changes without any warnings/announcement?

src/components/KeyboardToolbar/index.tsx
```diff
@@ -0,0 +1,149 @@
...
+
+import Arrow from "./Arrow";
+import Button from "./Button";
+import { colors } from "./colors";
+
+import type { KeyboardToolbarTheme } from "./colors";
+import type { ReactNode } from "react";
+
+export type KeyboardToolbarProps = {
+  content?: JSX.Element | null;
```
kirillzyusko, line 20: add JSDoc to all props

kirillzyusko:
Looks stable 😀

PR #350, opened 2024-02-01
feat: add enabled/disabled state for KeyboardAwareScrollView component

IvanIhnatsiuk:
## 📜 Description

I added the enableKeyboardHandling property to be able to turn off this component on screens that are no longer in focus and save resources. Also, it can be useful if someone uses bottomSheet navigation or has text input in a transparent modal or alerts.

## 💡 Motivation and Context

To give an ability for users to turn off auto scrolling (sometimes it may be needed to cover specific requirements).

## 📢 Changelog

- added new `enabled` prop for `KeyboardAwareScrollView` component;

Tested on:
- iPhone 15 pro
- Pixel 3 a

https://github.com/kirillzyusko/react-native-keyboard-controller/assets/86000012/584a63f0-cafc-4cf2-b207-6e4c33cbf86c

## 📝 Checklist

- [x] CI successfully passed


example/src/screens/Examples/AwareScrollView/index.tsx
```diff
@@ -33,6 +36,7 @@ export default function AwareScrollView({ navigation }: Props) {
     <KeyboardAwareScrollView
       testID="aware_scroll_view_container"
       bottomOffset={50}
+      enableKeyboardHandling={isFocused}
```
kirillzyusko, line 39: Maybe let's simply use `enabled` property?

src/components/KeyboardAwareScrollView/index.tsx
```diff
@@ -26,6 +26,7 @@ type KeyboardAwareScrollViewProps = {
   bottomOffset?: number;
   /** Prevents automatic scrolling of the `ScrollView` when the keyboard gets hidden, maintaining the current screen position. Default is `false`. */
   disableScrollOnKeyboardHide?: boolean;
+  enableKeyboardHandling?: boolean;
```
kirillzyusko, line 29: Can you also write JSDoc for this field?

And can you also mention this filed in the documentation (1.10.0 + latest version)?

src/components/KeyboardAwareScrollView/index.tsx
```diff
@@ -238,6 +244,10 @@ const KeyboardAwareScrollView = forwardRef<
         onMove: (e) => {
           "worklet";
 
+          if (!enableKeyboardHandling) {
+            return;
+          }
```
kirillzyusko, line 249: Wouldn't it be easier to add a check in `maybeScroll` function?

Because actually right now scrolling happens not only in `useKeyboardHandler`, but also in `useAnimatedReaction` (text input changes height) and in `onChangeText`, but in all these places we do call a `maybeScroll`. What do you think?

kirillzyusko (changes requested):

example/src/screens/Examples/AwareScrollView/index.tsx
```diff
@@ -15,6 +16,8 @@ export default function AwareScrollView({ navigation }: Props) {
   const [disableScrollOnKeyboardHide, setDisableScrollOnKeyboardHide] =
     useState(false);
 
+  const isFocused = useIsFocused();
```
kirillzyusko, line 19: Ideally would be good to add a button somewhere in header to be able to toggle this property and cover it by e2e tests 🙏

src/components/KeyboardAwareScrollView/index.tsx
```diff
@@ -75,6 +76,7 @@ const KeyboardAwareScrollView = forwardRef<
       children,
       bottomOffset = 0,
       disableScrollOnKeyboardHide = false,
+      enableKeyboardHandling = true,
```
kirillzyusko, line 79: Also, please add this property to snapshot tests (in example and FabricExample apps)

example/src/screens/Examples/AwareScrollView/index.tsx
IvanIhnatsiuk: we already have `enabled` prop in ScrollView, I think it's better to not override it

kirillzyusko:
> no longer in focus and save resources. Also, it can be useful if someone uses bottomSheet navigation or has text input in a transparent modal or alerts

I think this PR https://github.com/kirillzyusko/react-native-keyboard-controller/pull/347 will handle it automatically (it checks that input belongs to given ScrollView).

But having `enabled` property may still be handy in some corner cases (when we want to disable handling for a particular flow), so I'm not agains to have it 👀

src/components/KeyboardAwareScrollView/index.tsx
IvanIhnatsiuk: I think we can add an additional condition to the `maybeScroll` function to avoid changes when typing text. Also, I think we should leave this condition in `onStart`, `onEnd`, `onMove` events to avoid some calculations. What do you think?

kirillzyusko:
And what about bottom padding that applied automatically? Should we still apply it or not? What is the behavior of `react-native-keyboard-aware-scroll-view` in this case?

PR #380, opened 2024-03-08
fix: android build fail with rnx kit & treeShaking enabled

ashwin1014:
## 📜 Description

Fixed build issues when using `@rnx-kit`.

## 💡 Motivation and Context
When using tools like @rnx-kit from microsoft, which uses esbuild to handle bundling and also enables treeShaking, the build fails

## 📢 Changelog
- instead of direct export of type Ex: export { KeyboardToolbarProps } we export using "type": export { type KeyboardToolbarProps }

Screenshot 2024-03-07 at 23 01 07
Screenshot 2024-03-07 at 23 00 53

### DIFF
--- a/src/components/index.ts
+++ b/src/components/index.ts
@@ -4,5 +4,5 @@ export { default as KeyboardAwareScrollView } from "./KeyboardAwareScrollView";
 export {
   default as KeyboardToolbar,
   DefaultKeyboardToolbarTheme,
-  type KeyboardToolbarProps,
 } from "./KeyboardToolbar";
+export type { KeyboardToolbarProps } from "./KeyboardToolbar"
--- a/src/components/index.ts
+++ b/src/components/index.ts
@@ -5,4 +5,4 @@ export {
   default as KeyboardToolbar,
   DefaultKeyboardToolbarTheme,
 } from "./KeyboardToolbar";
-export type { KeyboardToolbarProps } from "./KeyboardToolbar"
+export type { KeyboardToolbarProps } from "./KeyboardToolbar";
--- a/src/components/index.ts
+++ b/src/components/index.ts
@@ -4,5 +4,5 @@ export { default as KeyboardAwareScrollView } from "./KeyboardAwareScrollView";
 export {
   default as KeyboardToolbar,
   DefaultKeyboardToolbarTheme,
-  KeyboardToolbarProps,
+  type KeyboardToolbarProps,
 } from "./KeyboardToolbar";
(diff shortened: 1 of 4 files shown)


src/components/index.ts
```diff
@@ -4,5 +4,5 @@ export { default as KeyboardAwareScrollView } from "./KeyboardAwareScrollView";
 export {
   default as KeyboardToolbar,
   DefaultKeyboardToolbarTheme,
-  KeyboardToolbarProps,
+  type KeyboardToolbarProps,
```
kirillzyusko, line 7: Can it be `export type { KeyboardToolbarProps } from "./KeyboardToolbar"` as I did in https://github.com/kirillzyusko/react-native-keyboard-controller/pull/299?

src/monkey-patch.android.ts
```diff
@@ -1,39 +1,40 @@
...
+  },
+};
 
-    setHidden(hidden: boolean): void {
-      RCTStatusBarManagerCompat.setHidden(hidden);
-    },
-  };
+// Define a function to apply the monkey patch
+export const applyMonkeyPatch = () => {
+  Object.assign(NativeAndroidManager.default, ModifiedNativeAndroidManager);
```
kirillzyusko, line 34: Thanks! I'll check how it works today and if everything is okay, then I'll merge it 👍

src/monkey-patch.android.ts
```diff
@@ -1,39 +1,40 @@
-// @ts-expect-error because there is no corresponding type definition
```
kirillzyusko, line 1: I think this line is needed - otherwise CI is failing 😔

src/components/index.ts
ashwin1014: yes

src/monkey-patch.android.ts
ashwin1014: added this

src/components/index.ts
ashwin1014: changed

src/monkey-patch.android.ts
ashwin1014: had a little help from gpt!

kirillzyusko:
@ashwin1014 you also need to run `yarn prettier '**/*' --write --ignore-unknown` to fix eslint stage

kirillzyusko:
@ashwin1014 by the way - did you check? Do these changes help to fix build issues?

ashwin1014:
> @ashwin1014 by the way - did you check? Do these changes help to fix build issues?

i was usually doing these changes manually from node_modules on my project. Then it used to build correctly

ashwin1014:
> @ashwin1014 you also need to run `yarn prettier '**/*' --write --ignore-unknown` to fix eslint stage
> 
> And there is one more eslint violation - please, fix it 🙏

Will do

kirillzyusko:
@ashwin1014 thank you for your contribution!

Issue #393, opened 2024-03-20
Runtime Error on iOS


arekkubaczkowski:
**Describe the bug**
I've installed the latest version of library and wrapped the app with Provider component which thrown the error right away. (See screenshots)
I confirm that removing provider fixes both errors.
I also tried to remove the entire components three from the app and the issue still exists.

<img width="607" alt="Screenshot 2024-03-20 at 07 57 15" src="https://github.com/kirillzyusko/react-native-keyboard-controller/assets/18654447/4f682266-17ea-4152-88ab-fc91628a22dd">

My tech stack is:
- Expo 50
- NX Monorepo

**Screenshots**
<img width="316" alt="Screenshot 2024-03-20 at 07 53 55" src="https://github.com/kirillzyusko/react-native-keyboard-controller/assets/18654447/a4b9f6d4-be6a-4675-8527-326eee459099">
<img width="313" alt="Screenshot 2024-03-20 at 07 54 19" src="https://github.com/kirillzyusko/react-native-keyboard-controller/assets/18654447/165387d5-5809-43a7-bb1f-1d1d0dc26b30">

**Smartphone (please complete the following information):**

- Desktop OS: [MacOS 14.1.2]
- Device: [iPhone 15]
- OS: [iOS 17.2]
- RN version: [0.73.5]
- RN architecture: [old]
- JS engine: [Hermes]
- Library version: [1.11.3]

**Additional context**

kirillzyusko:
@arekkubaczkowski can you try to use latest version of the library, for example 1.11.0?

arekkubaczkowski:
sorry, it was a typo, I am already using `1.11.3`

kirillzyusko:
@arekkubaczkowski okay👍May I ask you to share your `package.json`?

I think the same problem has been reported here https://github.com/kirillzyusko/react-native-keyboard-controller/issues/294

But that project used many private deps so I couldn't reproduce the problem.

If you can create a small reproduction example that would be a dope!

arekkubaczkowski:
as my app is under quite big monorepo there are also many dependencies:

```
 "devDependencies": {
        "@aws-sdk/client-s3": "^3.295.0",
        "@aws-sdk/credential-providers": "^3.299.0",
        "@babel/core": "^7.14.5",
        "@babel/preset-react": "^7.14.5",
        "@bahmutov/cypress-code-coverage": "^2.6.1",
        "@config-plugins/detox": "~6.0.0",
[...]

Issue #294, opened 2023-12-13
Invariant Violation: `new NativeEventEmitter()` requires a non-null argument


riamon-v:
**Describe the bug**
I just installed the package in my React Native project and wrapped my whole app into the `KeyboardProvider` provided.
The app can't seem to run afterwards.

The app is a bare React Native workflow implementing `expo-modules-core`

Here is a snippet of my package.json, maybe there is a conflict ?

```
"dependencies": {
    "@apollo/client": "^3.7.3",
    "@hookform/resolvers": "^2.8.8",
    "@notifee/react-native": "^7.8.0",
    "@react-native-async-storage/async-storage": "^1.17.7",
    "@react-native-clipboard/clipboard": "^1.11.1",
    "@react-native-community/blur": "^4.3.0",
    "@react-native-community/checkbox": "^0.5.9",
    "@react-native-community/datetimepicker": "^7.0.1",
    "@react-native-community/geolocation": "^3.0.1",
    "@react-native-community/netinfo": "^9.3.6",
    "@react-native-firebase/app": "^16.4.6",
    "@react-native-firebase/auth": "^16.4.6",
    "@react-native-firebase/crashlytics": "^16.4.6",
    "@react-native-firebase/messaging": "^16.4.6",
    "@react-native-firebase/perf": "^16.5.0",
    "@react-native-masked-view/masked-view": "^0.3.0",
    "@react-navigation/bottom-tabs": "^6.5.3",
    "@react-navigation/elements": "^1.3.18",
    "@react-navigation/native": "^6.1.2",
    "@react-navigation/stack": "^6.3.11",
    "@testing-library/dom": "^9.0.0",
    "@types/react": "^18",
    "@types/react-native-snap-carousel": "^3.8.5",
    "base-64": "^1.0.0",
    "creditcardutils": "^1.0.0",
    "date-fns": "^2.29.2",
    "deprecated-react-native-prop-types": "^4.0.0",
    "expo": "^48.0.0",
    "expo-barcode-scanner": "~12.3.2",
    "fetch-mock": "^9.11.0",
    "firebase": "^9.14.0",
    "geofire-common": "^6.0.0",
    "google-libphonenumber": "^3.2.32",
    "graphql": "^16.5.0",
    "i18n-js": "^4.2.2",
    "jwt-decode": "^3.1.2",
    "patch-package": "^6.4.7",
    "piano-analytics-js": "^6.6.0",
[...]

PR #395, opened 2024-03-20
fix: avoid deep workletization

kirillzyusko:
## 📜 Description

Avoid deep worletization of `react-native` package.

## 💡 Motivation and Context

It looks like in specific project setup babel plugin may try to serialize entire `react-native` package (see https://github.com/kirillzyusko/react-native-keyboard-controller/issues/393 for more context).

To avoid that in this PR I'm extracting `Platform.OS` into separate `OS` variable, so that babel plugin will capture only value of `OS` (and not entire `Platform` object).

Closes https://github.com/kirillzyusko/react-native-keyboard-controller/issues/393 and https://github.com/kirillzyusko/react-native-keyboard-controller/issues/294 

### JS

- extract `Platform.OS` into separate `OS` variable

Tested in the project where this error occurs.

## 📝 Checklist

- [x] CI successfully passed


kirillzyusko:
Thanks @arekkubaczkowski for catching and finding/testing the fix 😊

PR #397, opened 2024-03-21
chore: better sticky view keyboard interpolation

arekkubaczkowski:
### JS

-
-

### iOS

-
-

### Android

-
-

## 📝 Checklist

- [ ] CI successfully passed

### DIFF
--- a/src/components/KeyboardStickyView/index.tsx
+++ b/src/components/KeyboardStickyView/index.tsx
@@ -1,8 +1,5 @@
 import React, { forwardRef, useMemo } from "react";
-import Reanimated, {
-  interpolate,
-  useAnimatedStyle,
-} from "react-native-reanimated";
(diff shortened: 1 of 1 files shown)


kirillzyusko:
Hey @arekkubaczkowski 

This code introduces breaking changes which I'd like to avoid:

|Old|New|
|---|-----|
|image|image|

If we are using negative offset for `closed`, then it'll be ignored 😔 

Maybe we can add a new property, like `ignoreHeightBelow` and use interpolation like:

```tsx
interpolate(height.value, [0, ignoreHeightBelow, height.value], [closed, closed, opened])
```

Will it work for you?

arekkubaczkowski:
@kirillzyusko Nice spot, I had to miss that. Yeah, having an option to toggle those two behaviours makes sense for me! will push the changes later on.

Issue #398, opened 2024-03-21
KeyboardAvoidingView doesn't work as intended on Android 10


breakingrobot:
**Describe the bug**
On an Android 10 React Native app using a multi-step form with two screens that include an autofocused `TextInput`, any `router.navigate` or `router.push` will succeed but the following will happen

- While navigation from ScreenA to ScreenB, keyboard doesn't seem to be dismissed
- On ScreenB, TextInput is focused but the keyboard has no animation and `KeyboardAvoidingView` does not apply the padding expected. 

**Repo for reproducing**
https://github.com/breakingrobot/react-native-keyboard-controller/tree/android-10-bug

**To Reproduce**

1. Run an emulator with Android SDK <=29 (Android 10 and lower)
2. Build the example app from my fork 
3. Go to Example 14. Keyboard Avoiding View
4. Press Submit
4. Go back and retry

**Expected behavior**
Behavior that exhibit any Android devices with Android 11 and higher : 
- While navigating from ScreenA to ScreenB, keyboard should be dismissed
- On ScreenB, TextInput is focused, keyboard and `KeyboardAvoidingView` should trigger its animation and apply the correct padding

**Screenshots**
### Android 10 - SDK 29
https://github.com/kirillzyusko/react-native-keyboard-controller/assets/15056336/ff1dc4ee-350c-4770-92e3-56938bcc2c82

### Android 14 - SDK 34
https://github.com/kirillzyusko/react-native-keyboard-controller/assets/15056336/59bd57d0-6b80-4000-8f62-10089fc42429

**Smartphone (please complete the following information):**

- Desktop OS: macOS 14.0 (Sonoma)
- Device: 
  - Failing device: Emulated Pixel 4 - SDK 29 (arm64_v8a)
  - Successful device: Emulated Pixel 8 - SDK 34 (arm64_v8a)
- OS: Android 10 (SDK 29)
- RN version: 0.73.6 / 0.73.4 (on our Expo Project)
- RN architecture: Fabric
- JS engine: Hermes
- Library version: 1.11.4

kirillzyusko:
@breakingrobot Interesting! I thought it could be fixed in https://github.com/kirillzyusko/react-native-keyboard-controller/pull/391 but it looks like you are using latest version and if the issues is reproducible - it means that it's something else.

On your video I see, that on Android 14 keyboard starts to hide, but then appears again, while on Android 10 it always stays on the screen.

[...]

PR #399, opened 2024-03-25
fix: consume context values as a default one

kirillzyusko:
## 📜 Description

Consume context values (as default values).

## 💡 Motivation and Context

When we navigate between screens, sometimes there can be a situation, where keyboard is not moving. If we initialize values as a default one - we'll have incorrect layout, because keyboard is actually shown (and wasn't moved) so we will continue to use default values (keyboard closed) while keyboard is actually open.

To fix this problem I specify default values from the context - context always holds a current values. It fixes the problem 🙂 

Closes https://github.com/kirillzyusko/react-native-keyboard-controller/issues/398

### JS

- consume context values as a default one

Tested on Android 10.

|Before|After|
|-------|----|
|image|image|

## 📝 Checklist

- [x] CI successfully passed

### DIFF
--- a/src/components/KeyboardAvoidingView/hooks.ts
+++ b/src/components/KeyboardAvoidingView/hooks.ts
@@ -1,12 +1,16 @@
 import { useSharedValue } from "react-native-reanimated";
 
-import { useKeyboardHandler } from "react-native-keyboard-controller";
+import {
+  useKeyboardContext,
+  useKeyboardHandler,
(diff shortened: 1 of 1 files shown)


kirillzyusko:
@breakingrobot may I kindly ask you to test this PR? 🙏 

I believe it should fix the problem 😊

kirillzyusko:
@breakingrobot okay, thank you! I'll merge it and publish `1.11.5` soon 👀

Issue #307, opened 2023-12-23
useReanimatedKeyboardAnimation doesn't look smooth like the Reanimated UseAnimatedKeyboard hook in IOS when closing the keyboard


devoren:
**Describe the bug**
Hello! I'm creating a chat app and using fakeView. I'm using useReanimatedKeyboardAnimation but the height value is faster than useAnimatedKeyboard height value in iOS. Also how to add instets bottom? when i use useReanimatedKeyboardAnimation i use it like this: `height: height.value ? Math.abs(height.value) : insets.bottom` but it is not working with useAnimatedKeyboard (jumping)

**Code snippet**
Add your code snippet where error has been occurred.
```
<>
const { height } = useReanimatedKeyboardAnimation(); // ---- useAnimatedKeyboard ----

const fakeView = useAnimatedStyle(
() => ({
	height: height.value ? Math.abs(height.value) : insets.bottom,
}),[],);

<View style={styles.container}>
	<Chats.MessagesList
		data={allMessages}
	/>
	<Chats.MessageActions
		chat={chat}
		setMessages={setMessages}
	/>
	<Animated.View style={fakeView} />
      </View>
</>
```
**Repo for reproducing**
I would be highly appreciate if you can provide repository for reproducing your issue. It can significantly reduce the time for discovering and fixing the problem.
https://github.com/kirillzyusko/react-native-keyboard-controller/blob/main/example/src/screens/Examples/ReanimatedChatFlatlist/index.tsx

**To Reproduce**
1. Use ReanimatedChatFlatlist example
2. Open on ios simulator
3. Change useReanimatedKeyboardAnimation to useAnimatedKeyboard
4. Check the difference when opening/closing the keyboard.

**Expected behavior**
The keyboard closes smoothly, as when using the useAnimatedKeyboard hook.

**Screenshots**

**useReanimatedKeyboardAnimation:**

https://github.com/kirillzyusko/react-native-keyboard-controller/assets/99968085/c6514ea1-c8fd-4f5d-9687-ffa51bdf8d9f

**useAnimatedKeyboard:**

https://github.com/kirillzyusko/react-native-keyboard-controller/assets/99968085/21ba6bf7-dfa8-4a53-9f57-db41c9503bc5

**Smartphone (please complete the following information):**
 - Desktop OS: [MacOS]
 - Device: [Phone14]
 - OS: [iOS 16.6]
 - RN version: [0.72.6]
 - RN architecture: [old]
[...]

PR #400, opened 2024-03-25
fix: schedule missing layout animations

kirillzyusko:
## 📜 Description

Schedule layout animations on before keyboard appear/disappear.

## 💡 Motivation and Context

Looks like after RN 0.71 the disappear animation happens instantly if user pressed on the `ScrollView`. I don't know what was exactly broken, but I discovered, that if we schedule `LayoutAnimation` with `keyboard` type, then animation happens as it should.

So in this PR I decided to schedule layout animations manually and remove them when keyboard finished its movement.

Theoretically it's a little bit dangerous to schedule global animations, but:
- on iOS they are already kind of getting scheduled;
- we remove our animation after keyboard movement completed.

So it's kind of relatively safe 🙃 

Closes https://github.com/kirillzyusko/react-native-keyboard-controller/issues/307

### JS

- added `reanimated chat flat list` example to Fabric project;

### iOS

- added `scheduleKeyboardAnimation`/`unscheduleKeyboardAnimation` methods to `RCTUIManager` (as extension);
- added `onRequestAnimation`/`onCancelAnimation` methods to `KeyboardMovementObserver`;
- call `onRequestAnimation` on `keyboardWill` events and `onCancelAnimation` on `keyboardDid` events;

Tested on iPhone 15 Pro (both paper and fabric).

|Before|After|
|------|------|
|<video src="https://github.com/kirillzyusko/react-native-keyboard-controller/assets/22820318/f1a550e4-bb45-48e6-a4ee-7ef8a0bf1e0f">|<video src="https://github.com/kirillzyusko/react-native-keyboard-controller/assets/22820318/daafb826-df26-4dce-98af-d2d344f77699">|

## 📝 Checklist

- [x] CI successfully passed


focux:
I'm currently experiencing this as well, thank you so much for the fix.

Are these change already on NPM?

PR #419, opened 2024-04-22
feat: `KeyboardToolbar` callbacks

kirillzyusko:
## 📜 Description

Added `onDoneCallback`/`onNextCallback`/`onPrevCallback` functions to `KeyboardToolbar`.

## 💡 Motivation and Context

It's often needed to customize behavior when buttons are pressed, for example do a haptic feedback or even play a sound.

Before this PR it wasn't possible to add an action that will be called whenever user presses a particular button along with the default action. But in this PR I added it.

Initially I wanted to provide `onNextPress` functions that would redefine an entire handler. But in this case users will have to go into source code and understand how `KeyboardToolbar` works under the hood. Even though it's not difficult it's better to keep such things internally (and in most cases users will rely on defult behavior - in case if we need to re-define a behavior I'll add new functions later).

### Docs

- added docs about new `onDoneCallback`/`onNextCallback`/`onPrevCallback` callbacks;

### JS

- added `onDoneCallback`/`onNextCallback`/`onPrevCallback`;

Tested on Pixel 7 Pro (android 14) and iPhone 11 (iOS 17.4).

There is no way to test haptic visually 😀 

## 📝 Checklist

- [x] CI successfully passed

### DIFF
--- a/FabricExample/ios/Podfile.lock
+++ b/FabricExample/ios/Podfile.lock
@@ -1568,7 +1568,7 @@ EXTERNAL SOURCES:
 SPEC CHECKSUMS:
   boost: d3f49c53809116a5d38da093a8aa78bf551aed09
   CocoaAsyncSocket: 065fd1e645c7abab64f7a6a2007a48038fdc6a99
-  DoubleConversion: 76ab83afb40bddeeee456813d9c04f67f78771b5
+  DoubleConversion: fea03f2699887d960129cc54bba7e52542b6f953
   FBLazyVector: f64d1e2ea739b4d8f7e4740cde18089cd97fe864
   Flipper: c7a0093234c4bdd456e363f2f19b2e4b27652d44
   Flipper-Boost-iOSX: fd1e2b8cbef7e662a122412d7ac5f5bea715403c
--- a/FabricExample/ios/Podfile.lock
+++ b/FabricExample/ios/Podfile.lock
@@ -1273,6 +1273,13 @@ PODS:
     - ReactCommon/turbomodule/bridging
     - ReactCommon/turbomodule/core
     - Yoga
+  - RNReactNativeHapticFeedback (2.2.0):
+    - RCT-Folly
+    - RCTRequired
+    - RCTTypeSafety
+    - React-Codegen
+    - React-Core
+    - ReactCommon/turbomodule/core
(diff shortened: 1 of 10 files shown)


tmoran-stenoa:
@kirillzyusko Thank you for this, I just used this feature today (for haptics as you wrote)! I was thinking it could also be useful to be able to cancel the next/prev click if we want to handle the click in a custom way. I'm thinking of working on a PR for this, if you agree? This can be done with minimal changes, here's a quick draft example of the modified `onPressNext` handler:
```ts
const onPressNext = useCallback(() => {
    const event = new KeyboardToolbarArrowPressEvent("next");
    
    onNextCallback?.(event);
    if (!event.isCancelled) goToNextField();
}, [onNextCallback]);


// somewhere else in the code:
class KeyboardToolbarArrowPressEvent {
   isCancelled: boolean;
   constructor(public type: 'next' | 'prev') {}

   cancel() {
     this.isCancelled = false
   }
}
```

kirillzyusko:
Thanks @tmoran-stenoa for your comment!

Actually me and @IvanIhnatsiuk were thinking on how to handle that (we were discussing whether we should give an ability for user to fully implement logic of the handler (like providing `onPressNext` prop and if it's provided then call only this method and skip default implementation, but then decided to add corresponding `*Callback` methods).

In my opinion you are suggesting really great addition, but maybe instead of re-implementing event methods we can simply pass `e` from `Touchable*` to our method and then modify code like:

```tsx
const onPressNext = useCallback((e) => {    
    onNextCallback?.(e);
    if (!e.defaultPrevented) goToNextField();
}, [onNextCallback]);
```

and in our custom handler we can do:

```tsx
const onPressNextCallback = useCallback((e) => {
  if (shouldNotGoFurther) {
    e.preventDefault();
  }
}, []);
```

Curious to know what do you think about that 😊

tmoran-stenoa:
@kirillzyusko Good question. It's up to you, I don't think it makes much of a difference.

My initial instinct was that it would be better to not create a coupling between the toolbar's handling of the click event with React Native's event system, because the latter comes with a bunch of baggage that may interfere with our purposes and may cause confusion.

Anyhow, I think both ways should work. Let me know which way you prefer and I can work on a PR if you want.

Issue #420, opened 2024-04-23
`KeyboardAwareScrollView` is not reacting on text selection 


kirillzyusko:
**Describe the bug**

If your multiline input is partially covered by the keyboard and you select text inside this input, then selection can go under the keyboard and you will not understand what you are selecting.

**Code snippet**

`KeyboardAwareScrollView` example.

**Repo for reproducing**

Can be reproduced in example app.

**To Reproduce**

1. Go to `KeyboardAwareScrollView` example
2. Click on TextInput#5
3. Type 10 lines of text
4. Scroll down, to make visible only one line of the text
5. Try to select an entire text

**Expected behavior**

On web for example we have following behavior:

https://github.com/kirillzyusko/react-native-keyboard-controller/assets/22820318/ab82f7fe-8bf9-4ec9-9f3b-41e642a3a7bc

**Screenshots**

|iOS      |Android|
|-------|--------|
|<video src="https://github.com/kirillzyusko/react-native-keyboard-controller/assets/22820318/42acc001-64fe-4fcf-955e-dd2ea20fe244">|<video src="https://github.com/kirillzyusko/react-native-keyboard-controller/assets/22820318/88834cf9-d14e-4f2c-b4c2-68066fc5902e">|

**Smartphone (please complete the following information):**

- Desktop OS: macOS 14.4.1
- Device: iPhone 11 (iOS 17.4)/Pixel 7 Pro (Android 14)
- OS: see above
- RN version: 0.73.6
- RN architecture: old/new
- JS engine: Hermes
- Library version: 1.11.6

**Additional context**

I think this library should expose additional handler in `useFocusedInputHandler`:

```tsx
useFocusedInputHandler({
  onSelectionChange: (e) => {
    "worklet";
  }
});
```

That will be fired whenever position of the carret will be changed. Also probably we'll need to include `x`/`y` coordinates for first and last carret.

PR #426, opened 2024-04-25
feat: onSelectionChange

kirillzyusko:
## 📜 Description

Added a new `onSelectionChange` handler for `useFocusedInputHandler` hook.

## 💡 Motivation and Context

These changes will be a findamental work for resolving https://github.com/kirillzyusko/react-native-keyboard-controller/issues/420

To get `onSelectionChange` events I had to intercept them:
- on iOS it's possible on a delegate level (since default delegate forwards all events to JS I had to carefully intercept all events and forward them to original delegate - that's why I create `KCTextInputCompositeDelegate` (KC stands for `KeyboardController`);
- on Android I can inject my own handler, but turned out that on Fabric hanlder is re-created on each symbol typing, so my handler gets overwritten by handler from props - that's why I added `FrameScheduler` class and went with "long pooling" approach (later on I plan to add `addSelectionListener` method to RN core);

Last but not least - interaction with the proxy-delegate will give us an ability to intercept more events, so we may get rid off patch for `react-native-text-input-mask` and get events directly from the delegate.

### Docs

- added a reference about new handler and its use cases;

### E2E

- updated e2e test suite (covered new functionality);

### JS

- added new event tye declaration;
- added new internal hook (`useFocusedInputSelectionHandler`);
- added new unit tests;
- updated fabric specs;
- update example app;
- updated `cspell` config;

### iOS

- added `KCTextInputCompositeDelegate`;
- added `setForceDelegate` extension for `RCTUITextView`;
- added new event;

### Android

- added `FrameScheduler` class;
- added extension for observing selection changes;
- added new event;

Tested on:
- Pixel 7 Pro (Android 14, real device);
- Pixel 3A (Androd 13, emulator);
- iPhone 15 Pro (simulator, iOS 17.4);
- iPhone 11 (iOS 17.4.1, real devcie).

## 📝 Checklist

- [x] CI successfully passed


src/reanimated.native.ts
```diff
@@ -82,24 +86,51 @@ export const useFocusedInputLayoutHandler: FocusedInputLayoutHandlerHook<
...
+
       const { onFocusedInputTextChanged } = handlers;
 
       if (
         onFocusedInputTextChanged &&
         event.eventName.endsWith("onFocusedInputTextChanged")
       ) {
-        onFocusedInputTextChanged(event, context);
+        onFocusedInputTextChanged(
+          event as FocusedInputTextChangedEvent,
```
kirillzyusko, line 105: Should be reverted

ios/delegates/KCTextInputCompositeDelegate.swift
```diff
@@ -0,0 +1,124 @@
...
+            "y": selection.startY,
+            "position": selection.start,
+          ],
+          "end": [
+            "x": selection.endX,
+            "y": selection.endY,
+            "position": selection.end,
+          ],
+        ],
+        "target": UIResponder.current.reactViewTag,
```
kirillzyusko, line 120: Maybe better would be to inject `target` on `FocusedInputObserver` level?

Issue #450, opened 2024-05-20
snapToOffsets / snapToInterval if keyboard moves scrollview (KeyboardAwareScrollView)


kirillzyusko:
### Discussed in https://github.com/kirillzyusko/react-native-keyboard-controller/discussions/449

<div type='discussions-op-text'>

<sup>Originally posted by **kgni** May 17, 2024</sup>
Hi! 

Not really sure where to put a question like this. If this is the wrong place, please let me know and I will repost my questions somewhere else :)

I have the following layout, where I have a `header` and a `stickyHeader`.

### Header:

![CleanShot 2024-05-17 at 21 12 29](https://github.com/kirillzyusko/react-native-keyboard-controller/assets/84397151/d352246b-e082-4d01-9d8f-bcac2be52cdb)

### Sticky Header (scrolled)
![CleanShot 2024-05-17 at 21 12 56](https://github.com/kirillzyusko/react-native-keyboard-controller/assets/84397151/fa3a1712-8c86-4b39-a909-d0d71048ad19)

When scrolling the header will disappear and I also have `snapToOffsets` set for when you are scrolling manually (not the keyboard scrolling), to ensure that I don't have any states in between the header being shown and hidden - E.G both the sticky header title and the normal header title would be shown at the same time.

What I'm having trouble with, is when I have an input that only causes the keyboard to scroll a tiny bit, instead of snapping to the interval to prevent multiple "states" appearing at the same time:

![CleanShot 2024-05-17 at 21 10 42](https://github.com/kirillzyusko/react-native-keyboard-controller/assets/84397151/bed08178-9002-4622-969f-ae9337cf8267)

Essentially what I want to do, is to ensure that the keyboard as a minimum is scrolling the scrollview to the `snapToOffset` / `snapToInterval` position

Does anyone have an idea of how I can implement this?

Thanks in advance!

Here is my code for reference:

```javascript

import CONSTANTS from '@constants';
import { usePadding, useScreenSize } from '@hooks';
import { useRouter } from 'expo-router';
import { StatusBar } from 'expo-status-bar';
import { forwardRef, useCallback, useEffect, useState } from 'react';
import {
  LayoutChangeEvent,
  LayoutRectangle,
  ScrollViewProps,
  View,
} from 'react-native';
import {
  KeyboardAvoidingView,
  KeyboardAwareScrollView, useReanimatedKeyboardAnimation
} from 'react-native-keyboard-controller';
import Animated, {
[...]

PR #452, opened 2024-05-23
feat: scroll to at least the first snapToOffset value if any

kgni:
## 💡 Motivation and Context

When dealing with Scrollviews that are snapping to a specific point, if the keyboard is creating a scroll, we in some cases want the scrollview to scroll to at least the first snapToOffsets

It fixes layouts being stuck between 2 states, please see the following issue for more info: https://github.com/kirillzyusko/react-native-keyboard-controller/issues/450

Closes https://github.com/kirillzyusko/react-native-keyboard-controller/issues/450

### JS
- Added `clampLower` utility function
- Use this `clampLower` function to ensure a minimum scroll to the first snapToOffset if any

Tested really briefly, going to point my own project to this PR for further testing

To be added

## 📝 Checklist

- [ ] CI successfully passed

### DIFF
--- a/example/src/screens/Examples/AwareScrollView/index.tsx
+++ b/example/src/screens/Examples/AwareScrollView/index.tsx
@@ -23,6 +23,8 @@ export default function AwareScrollView({ navigation }: Props) {
     useState(false);
   const [enabled, setEnabled] = useState(true);
 
+  const [snapToOffsetsEnabled, setSnapToOffsetsEnabled] = useState(false);
+
   useEffect(() => {
     navigation.setOptions({
       headerRight: () => (
@@ -37,16 +39,52 @@ export default function AwareScrollView({ navigation }: Props) {
     });
   }, []);
 
+  const snapToOffsets = [125, 225, 325, 425];
+
   return (
     <>
       <KeyboardAwareScrollView
         testID="aware_scroll_view_container"
         bottomOffset={50}
         enabled={enabled}
+        snapToOffsets={snapToOffsetsEnabled ? snapToOffsets : undefined}
         disableScrollOnKeyboardHide={disableScrollOnKeyboardHide}
         style={styles.container}
         contentContainerStyle={styles.content}
       >
+        {snapToOffsetsEnabled && (
+          <>
+            {snapToOffsets.map((offset) => (
+              <View
+                key={offset}
+                style={{
+                  width: "100%",
+                  position: "absolute",
+                  top: offset,
+                }}
+              >
+                <View
+                  style={{
+                    flexDirection: "row",
+                    alignItems: "center",
+                    gap: 12,
+                  }}
+                >
+                  <Text>{offset}</Text>
+                  <View
+                    key={offset}
+                    style={{
+                      height: 2,
+                      flex: 1,
+                      backgroundColor: "black",
+                    }}
+                  ></View>
+                </View>
+              </View>
+            ))}
+          </>
+        )}
+
         {new Array(10).fill(0).map((_, i) => (
           <TextInput
             key={i}
@@ -78,6 +116,17 @@ export default function AwareScrollView({ navigation }: Props) {
             testID="bottom_sheet_toggle_enabled_state"
             onChange={() => {
               setEnabled(!enabled);
+            }}
+          />
+        </View>
+
+        <View style={styles.switchContainer}>
+          <Text>Toggle snapToOffsets</Text>
+          <Switch
+            value={snapToOffsetsEnabled}
+            testID="bottom_sheet_toggle_enabled_state"
+            onChange={() => {
+              setSnapToOffsetsEnabled(!snapToOffsetsEnabled);
             }}
           />
         </View>
--- a/example/src/screens/Examples/AwareScrollView/index.tsx
+++ b/example/src/screens/Examples/AwareScrollView/index.tsx
@@ -11,6 +11,7 @@ import type { ExamplesStackParamList } from "../../../navigation/ExamplesStack";
 import type { StackScreenProps } from "@react-navigation/stack";
 
 type Props = StackScreenProps<ExamplesStackParamList>;
+const snapToOffsets = [125, 225, 325, 425];
 
 export default function AwareScrollView({ navigation }: Props) {
   const bottomSheetModalRef = useRef<BottomSheet>(null);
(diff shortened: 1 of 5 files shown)


kgni:
https://github.com/kirillzyusko/react-native-keyboard-controller/assets/84397151/699d2ed7-1318-4b86-a261-ea1c613dcb71

kgni:
The only issue with this implementation currently, is that we have to ensure that the body (from the 1. Room address & name title -> Next button sticky footer), has a minHeight that is tall enough so that the keyboard can actually scroll to the snapToOffsets.

A side effect of that, is that the body for this layout is always scrollable to at least the snapToOffset, which is probably not desired. Instead the layout should not be scrollable whenever there is no need to.

I tried to create an implementation where the body would initially not be scrollable, by calculating the different elements height on initial Layout:

`screenHeight - headerHeight - footerHeight = bodyHeight`

My thought then was, that when the keyboard opens we would add to the body height to make it scrollable - however I ran into an issue where the keyboard would not trigger the scroll,  because the addition to the bodyHeight is done after the keyboard triggers (I used the useKeyboardHandler, to add the height of the keyboard to the minBodyHeight).

Do you have an idea how I could add to the bodyHeight and then trigger the scroll? @kirillzyusko 

Also if I'm not making sense, please let me know. I can create a loom explaining it more

kirillzyusko:
> I can create a loom explaining it more

Well, that would be good - because you are the only one person who has an access to the source code and for me it's slightly difficult to understand a references without an actual reproduction demo. So if you can create code sample and pos it here - that would be amazing!

kgni:
> So if you can create code sample and post it here - that would be amazing!

Sure thing!

I will probably create one by the end of the week

src/components/KeyboardAwareScrollView/index.tsx
```diff
@@ -153,11 +153,32 @@ const KeyboardAwareScrollView = forwardRef<
         const inputHeight = layout.value?.layout.height || 0;
         const point = absoluteY + inputHeight;
 
+        const scrollOutput = (
```
kirillzyusko, line 156: Can this function be moved into separate file?

This component already has pretty big amount of lines of code, so let's reduce it 😎 (especially we can do that, since it's pure function)

example/src/screens/Examples/AwareScrollView/index.tsx
```diff
@@ -37,16 +39,52 @@ export default function AwareScrollView({ navigation }: Props) {
...
+                >
+                  <Text>{offset}</Text>
+                  <View
+                    key={offset}
+                    style={{
+                      height: 2,
+                      flex: 1,
+                      backgroundColor: "black",
+                    }}
+                  ></View>
```
kirillzyusko, line 81: I think linter/prettier will comply bout this, but do not worries, I'll approve this PR and it will run CI - we can see what failed and what not 👀

example/src/screens/Examples/AwareScrollView/index.tsx
```diff
@@ -37,16 +39,52 @@ export default function AwareScrollView({ navigation }: Props) {
...
+                    gap: 12,
+                  }}
+                >
+                  <Text>{offset}</Text>
+                  <View
+                    key={offset}
+                    style={{
+                      height: 2,
+                      flex: 1,
+                      backgroundColor: "black",
```
kirillzyusko, line 79: Can this style be moved to static styles?

example/src/screens/Examples/AwareScrollView/index.tsx
```diff
@@ -37,16 +39,52 @@ export default function AwareScrollView({ navigation }: Props) {
...
+                  width: "100%",
+                  position: "absolute",
+                  top: offset,
+                }}
+              >
+                <View
+                  style={{
+                    flexDirection: "row",
+                    alignItems: "center",
+                    gap: 12,
```
kirillzyusko, line 70: Can this style be moved to static styles?

example/src/screens/Examples/AwareScrollView/index.tsx
```diff
@@ -37,16 +39,52 @@ export default function AwareScrollView({ navigation }: Props) {
     });
   }, []);
 
+  const snapToOffsets = [125, 225, 325, 425];
```
kirillzyusko, line 42: Let's move this variable outside of the function (in this case it will not be re-created on every render and it'll improve performance)

example/src/screens/Examples/AwareScrollView/index.tsx
kirillzyusko: Okay, it's not complaining, but can you use `<View />` syntax, please? 🙏

kirillzyusko:
Size diff job is failing - it's expected, because of GitHub security preferences (it can not send a comment to PR from external contributors - I haven't figured out how to fix it yet). So ignore it 🙃 

Size diff from this PR is 2kb.

example/src/screens/Examples/AwareScrollView/index.tsx
```diff
@@ -81,6 +119,17 @@ export default function AwareScrollView({ navigation }: Props) {
             }}
           />
         </View>
+
+        <View style={styles.switchContainer}>
+          <Text>Toggle snapToOffsets</Text>
+          <Switch
+            value={snapToOffsetsEnabled}
+            testID="bottom_sheet_toggle_enabled_state"
```
kirillzyusko, line 127: Please, change this `testId` (I think you copied/pasted it) - it causes e2e test to fail, because we have two identical ids and we expect it'll be only one 😓

example/src/screens/Examples/AwareScrollView/index.tsx
```diff
@@ -37,16 +39,52 @@ export default function AwareScrollView({ navigation }: Props) {
...
+        snapToOffsets={snapToOffsetsEnabled ? snapToOffsets : undefined}
         disableScrollOnKeyboardHide={disableScrollOnKeyboardHide}
         style={styles.container}
         contentContainerStyle={styles.content}
       >
+        {snapToOffsetsEnabled && (
+          <>
+            {snapToOffsets.map((offset) => (
+              <View
+                key={offset}
```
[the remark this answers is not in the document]
kgni: hmm, not in my IDE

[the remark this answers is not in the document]
kgni: but not sure why I even had a key on the nested View, I will remove it

Issue #387, opened 2024-03-15
Opening Modal prevents proper resizing of layout


krystianhub:
**Describe the bug**
When the keyboard is open, launching a Modal (from either the react-native or react-native-modal package) results in the keyboard closing but without correctly adjusting the layout size.

**Code snippet**
```javascript
import { useState } from 'react';
import { Button,  FlatList,  StyleSheet, Text, TextInput, View } from 'react-native';
import { KeyboardAwareScrollView, KeyboardProvider } from 'react-native-keyboard-controller';
import Modal from 'react-native-modal';

const data = Array.from(Array(50).keys());

export default function App() {
  const [modalVisible, setModalVisible] = useState(false);

  const onPress = () => {
    // Partial workaround: uncomment me and set Modal coverScreen to false
    // KeyboardController.dismiss();
    setModalVisible(true);
  };

  return (
    <KeyboardProvider>

      <Modal isVisible={modalVisible} coverScreen={true}>
        <View style={{flex: 1, justifyContent: 'center', alignItems: 'center', backgroundColor: 'black'}}>
          <Text style={{color: 'white'}}>Modal content here</Text>

          <Button title='Close modal' onPress={() => setModalVisible(false)} />
        </View>
      </Modal>

      <KeyboardAwareScrollView contentContainerStyle={styles.container} keyboardShouldPersistTaps={"always"}>

        <View style={{flex: 1}} />

        <Button title="Open modal"  onPress={onPress}/>


        <FlatList scrollEnabled={false} data={data} style={{height: 500}} renderItem={({index}) => <Text style={{textAlign: 'center'}}>{index}</Text>}/>

        <TextInput placeholder='text input' style={{backgroundColor: 'white', padding: 10}} />

        <View style={{flex: 1}} />

      </KeyboardAwareScrollView>
    </KeyboardProvider>
  );
}

const styles = StyleSheet.create({
  container: {
    flex: 1,
    backgroundColor: 'white',
    alignItems: 'center',
    justifyContent: 'center',
    padding: 50,
  },
});

``` 

**Repo for reproducing**
https://github.com/krystianhub/RN-Modal-Issue - Expo Managed workflow; requires compiled app with expo-dev-client. 
I can provide the apk if needed. Alternatively, to build it locally yourself:
```shell
eas build --profile development --platform android --local
``` 

**To Reproduce**

1. Tap on text input element
2. Keyboard is open
3. Tap on open modal
[...]

Issue #369, opened 2024-02-25
Animated keyboard values do not update if the focussed TextInput exists within a modal


Braden1996:
**Describe the bug**
Animated keyboard values do not update if the focussed TextInput exists within a modal.

**Code snippet**
**Repo for reproducing**
Here is a test case I've been using which plugs into your example app:
```tsx
import React, { useState } from "react";
import { Button, Modal, StyleSheet, View } from "react-native";
import {
  GestureHandlerRootView,
  TouchableWithoutFeedback,
} from "react-native-gesture-handler";
import { KeyboardProvider } from "react-native-keyboard-controller";

import TextInput from "../../../components/TextInput";
import KeyboardAnimation from "../KeyboardAnimation";

export default function ModalExample() {
  const [modalVisible, setModalVisible] = useState(false);

  return (
    <View style={styles.container}>
      <Button
        title={"Show Modal"}
        onPress={() => setModalVisible(true)}
        testID="toggle_button"
      />

      <View style={styles.animationContainer}>
        <KeyboardAnimation />
      </View>

      <Modal
        transparent
        visible={modalVisible}
        onDismiss={() => setModalVisible(false)}
        onRequestClose={() => setModalVisible(false)}
      >
        <GestureHandlerRootView style={styles.modalContainer}>
            <TouchableWithoutFeedback onPress={() => setModalVisible(false)}>
              <TextInput
                testID="modal_example_text_input"
                style={styles.textInput}
              />
            </TouchableWithoutFeedback>
        </GestureHandlerRootView>
      </Modal>
    </View>
  );
}

const styles = StyleSheet.create({
  container: {
    flex: 1,
    justifyContent: "center",
    alignItems: "stretch",
  },
  modalContainer: {
    flex: 1,
    backgroundColor: "rgba(0,0,0,0.2)",
    justifyContent: "center",
    alignItems: "center",
  },
  animationContainer: {
    flex: 1,
  },
  textInput: {
    width: 200,
    marginTop: 50,
    height: 50,
    backgroundColor: "yellow",
  },
});
```

**To Reproduce**

1. Create a `TextInput` element inside of a React Native [Modal](https://reactnative.dev/docs/modal).
2. See that any keyboard height observers no longer seem to receive updates.

**Expected behavior**
RN Keyboard Controller's utilities should continue to work for TextInputs inside Modals.

**Screenshots**

[...]

PR #466, opened 2024-06-10
fix: modal integration

kirillzyusko:
## 📜 Description

Fixed a problem of keyboard movement not being detected in Modal window on Android.

## 💡 Motivation and Context

The most challenging in this PR is getting an access to `dialog`. Initially I thought to create an additional view (`ModalKeyboardProvider`) and then through this view get an access to the `dialog`, but after some experiments I realized, that we'll get an access to `DialogRootViewGroup` and this class is not holding a reference to `dialog`, so this idea failed.

Another approach was to listen to all events in `eventDispatcher` and when we detect `onShow` (`topShow`) event, then using `uiManager` we can resolve the view and get `ReactModalHostView`. This approach work, but it has one downside - we listen to all events. It doesn't hit the performance, but this approach looks like a workaround (it has too many transitive dependencies).

A PR https://github.com/facebook/react-native/pull/45534 introduces a new API for detection show/hide modals. I hope it'll be merged at some point of time and we can start to use new API. In a meantime I'll use the current approach and will incapsulate all work in `ModalAttachedWatcher` - in this case we have our own interface for dealing with modals, and we can change internals without affecting other files 😎

Closes https://github.com/kirillzyusko/react-native-keyboard-controller/issues/369 

Potentially helps to fix https://github.com/kirillzyusko/react-native-keyboard-controller/issues/387

### E2E

- added new `modal` test;

### Android

- added `ModalAttachedWatcher` class;
- `KeyboardAnimationCallback` now accepts config to reduce amount of passed params (`KeyboardAnimationCallbackConfig`);
- pass `eventPropagationView` to `KeyboardAnimationCallback` and `FocusedInputObserver`;
- return `WindowInsetsCompat.CONSUMED` from `onApplyWindowInsets` instead of `insets`;
- add `syncKeyboardPosition` to `KeyboardAnimationCallback`;
- dispatch the same events through cycle instead of code duplication.

Tested manually on:
- Xiaomi Redmi Note 5 Pro (Android 9);
- Pixel 7 Pro (Android 14);
- Pixel 3A (Android 13, emulator);
- Pixel 2 (AOSP, Android 9, e2e tests);
- Pixel 2 (Android 11, emulator, e2e tests).

|Before|After|
|-------|-----|
|Screenshot 2024-07-24 at 16 43 20|Screenshot 2024-07-24 at 16 41 39|

## 📝 Checklist

- [x] CI successfully passed


android/src/main/java/com/reactnativekeyboardcontroller/views/EdgeToEdgeReactViewGroup.kt
```diff
@@ -148,7 +163,7 @@ class EdgeToEdgeReactViewGroup(private val reactContext: ThemedReactContext) : R
     Handler(Looper.getMainLooper()).post { view.removeSelf() }
   }
 
-  private fun getContentView(): FitWindowsLinearLayout? {
+  private fun getContentView(): LinearLayout? {
```
kirillzyusko, line 166: Because `FitWindowsLinearLayout` can only be accessed from within the same library group prefix (referenced `groupId=androidx.appcompat` with prefix `androidx` from `groupId=...`) More... (⌘F1)

PR #500, opened 2024-07-08
feat: allow js and reanimated scroll handlers for `KeyboardAwareScrollView`

kirillzyusko:
## 📜 Description

Use `useScrollViewOffset` hook to get a scrolled distance.

## 💡 Motivation and Context

`useScrollViewOffset` hook doesn't substitute `onScroll` handlers with `dummyListener` and thus it allows you to define your custom `onScroll` handler (and at the same time it can be plain js handler or reanimated one).

The new approach gives us next advantages:
- `position.value` is still calculating on UI;
- you can pass both worklet/js-handler as `onScroll` and it will work out of the box without additional manipulations 😎
- we don't use `@ts-expect-error` and don't ship the code with potential errors;
- we can support reanimated up to `2.11.0` (the first release that added `useScrollViewOffset`).

The disadvantage is that we need to bump REA from `2.3.0` to `2.11.0` (I described that in documentation). Theoretically we can create our own version of `useScrollViewOffset` and support REA from `2.3.0` again. But version `2.11.0` was released 2 years ago:

image

And after math calculations I figured out that dropping support from `2.3.0` to `2.11.0` we'll loose only 2% of users (overall downloads for `2.3.0`-`2.10.0` are ~17k with 770k of overall downloads), so I don't think it's a breaking change.

And I think this 2% of users typically don't update deps at all in the project so I think it's not critical to bump minimal supported version for v1.12.x of this library.

### JS

- use `useScrollViewOffset` hook to get a scrolled distance;

Tested manually + on CI.

## 📝 Checklist

- [x] CI successfully passed

### DIFF
--- a/docs/docs/guides/compatibility.md
+++ b/docs/docs/guides/compatibility.md
@@ -13,7 +13,7 @@ If you found an incompatibility or conflict with other open source libraries - d
 
 :::
 
-## React Native
+## `react-native`
 
 Below you can find an information about compatibility with `react-native` package per different architectures.
 
@@ -25,6 +25,10 @@ Below you can find a table with supported versions:
 
 | library version | react-native version |
 | --------------- | -------------------- |
+| 1.12.0+         | 0.74.0+              |
+| 1.10.0+         | 0.73.0+              |
+| 1.6.0+          | 0.72.0+              |
+| 1.5.0+          | 0.71.0+              |
 | 1.3.0+          | 0.70.0+              |
 | 1.2.0+          | 0.69.0+              |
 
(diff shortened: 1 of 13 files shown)


kirillzyusko:
Hey @mobily

Can you check if this PR fixes the problem that we've been discussing in https://github.com/kirillzyusko/react-native-keyboard-controller/pull/490#issuecomment-2213163270?

I think it's the best solution, because:
- we don't use newly introduced API in reanimated (`useComposedEventHandlers`) and we can support pretty old reanimated versions;
- `position.value` is still calculating on UI;
- you can pass both worklet/js-handler as `onScroll` and it will work out of the box without additional manipulations 😎 
- we don't use `@ts-expect-error` and don't ship the code with potential errors

kirillzyusko:
@mobily did you have a chance to check that? 👀

mobily:
@kirillzyusko yup, I checked it out and it works like a charm! thanks a lot! ❤️

Issue #526, opened 2024-07-29
using  StatusBar.setTranslucent with the keyboardProvider 


AbdulkarimTaha:
Hi there , 
I previously had the ability to use StatusBar.setTranslucent or the StatusBar component on different pages to change the behavior of the status bar. However, when using the KeyboardProvider, I seem to have lost this ability. Now, I can only specify it in the <KeyboardProvider statusBarTranslucent={true}>. Even if I try to use StatusBar.setTranslucent(false) later, it does not change anything.

Is there a way to dynamically change the statusBarTranslucent property after initializing the KeyboardProvider?

kirillzyusko:
Thank you for the issue @AbdulkarimTaha 

It happen because of: https://github.com/kirillzyusko/react-native-keyboard-controller/blob/08efc819fcc6ae09dfb857b2ba6ee9bf17fa10d1/android/src/main/java/com/reactnativekeyboardcontroller/modules/StatusBarManagerCompatModuleImpl.kt#L55-L65

May I ask you to show a video (without KeyboardProvider) how transition in your app between translucent/non-translucent status bar looks like?

This is a "known" bug in the library - I just want to see a video confirmation, that there is indeed a use-case in a real world app (that uses both modes) 👀

PR #15, opened 2022-05-10
issue #14: translucent status bar on android

kirillzyusko:
Fixes: https://github.com/kirillzyusko/react-native-keyboard-controller/issues/14

This PR adds new prop called `statusBarTranslucent`

The problem was in the fact that in `ApplyWindowInsetsListener` we always add constant paddings. In case of translucent status bar we should apply 0 paddings. Prop `statusBarTranslucent` now control this aspect.

PR #528, opened 2024-07-31
feat: react on `StatusBar.translucent` changes

kirillzyusko:
## 📜 Description

Give an ability to control `translucent` from `StatusBar` component.

## 💡 Motivation and Context

The original decision to ignore `translucent` property was introduced in https://github.com/kirillzyusko/react-native-keyboard-controller/pull/30

The problem is that we already have `translucent` `StatusBar` and we can only mimicrate `non-translucent` by adding additional margin. The solution with static `statusBarTranslucent` property was added https://github.com/kirillzyusko/react-native-keyboard-controller/pull/15 and till now we didn't have an ability to manage `translucent` property dynamically

The new functionality was requested in https://github.com/kirillzyusko/react-native-keyboard-controller/issues/526

To implement new functionality I am discovering `edge-to-edge` view (managed by `KeyboardProvider`) and conditionally add remove padding, re-setup listener and re-request insets update.

Closes https://github.com/kirillzyusko/react-native-keyboard-controller/issues/526

### JS

- add `vendor` to prettier ignore;
- add `StatusBar` to root screen;
- add new dynamic props to StatusBar example (`translucent`) + module management;
- add testID to StatusBar example screen;

### E2E

- updated assets on Android;

### Android

- added new `ReactContext` extension file and move `.rootView` and `.content` extensions there;
- implemented `setTranslucent` method in `StatusBarCompatModule`;
- assign `TAG` to `EdgeToEdgeView`;

Tested manually on:
- Pixel 3a (API 33, emulator);
- Pixel 7 Pro (API 34, real device);
- (Xiaomi Redmi Note 5 Pro, API 28);

|Before|After|
|-------|-----|
|<video src="https://github.com/user-attachments/assets/e2b5e28f-6038-4d09-8c9e-2ef7d5015395">|<video src="https://github.com/user-attachments/assets/e089f3a4-f2bd-4e4a-9d32-0804c88ed2f4">|

## 📝 Checklist

- [x] CI successfully passed


android/src/main/java/com/reactnativekeyboardcontroller/views/EdgeToEdgeReactViewGroup.kt
```diff
@@ -200,4 +202,20 @@ class EdgeToEdgeReactViewGroup(private val reactContext: ThemedReactContext) : R
     }
   }
   // endregion
+
+  // region external methods
+  fun forceStatusBarTranslucent(isStatusBarTranslucent: Boolean) {
+    UiThreadUtil.runOnUiThread {
+      if (active) {
```
kirillzyusko, line 209: I think we need to change here only if `this.isStatusBarTranslucent != isStatusBarTranslucent`?

android/src/main/java/com/reactnativekeyboardcontroller/extensions/ThemedReactContext.kt
```diff
@@ -12,7 +12,7 @@ import com.facebook.react.uimanager.events.Event
 import com.facebook.react.uimanager.events.EventDispatcher
 import com.reactnativekeyboardcontroller.listeners.WindowDimensionListener
 
-val ThemedReactContext.rootView: View?
+val ReactContext.rootView: View?
```
kirillzyusko, line 15: Should it be moved to separate file now?..

android/src/main/java/com/reactnativekeyboardcontroller/views/EdgeToEdgeReactViewGroup.kt
```diff
@@ -200,4 +202,20 @@ class EdgeToEdgeReactViewGroup(private val reactContext: ThemedReactContext) : R
     }
   }
   // endregion
+
+  // region external methods
+  fun forceStatusBarTranslucent(isStatusBarTranslucent: Boolean) {
+    UiThreadUtil.runOnUiThread {
```
kirillzyusko, line 208: Maybe we should run it from UI thread from module package?
Component shouldn't know anything abouth the thread in this situation 🤔

PR #585, opened 2024-09-13
fix: increase minimum ios deployment to 12

matinzd:
## 📜 Description

Increase ios minimum deployment to 12

## 💡 Motivation and Context

When new arch is enabled xcode will complain with the following error in the latest RN version 0.75.3:

```
error: 'value' is unavailable: introduced in iOS 12.0
```

Screenshot 2024-09-13 at 14 20 16

### iOS

- Increase ios minimum deployment to 12

Run pod install and build again in xcode.

## 📝 Checklist

- [ ] CI successfully passed

### DIFF
--- a/react-native-keyboard-controller.podspec
+++ b/react-native-keyboard-controller.podspec
@@ -14,7 +14,7 @@ Pod::Spec.new do |s|
   s.license      = package["license"]
   s.authors      = package["author"]
 
-  s.platforms    = { :ios => "11.0" }
+  s.platforms    = { :ios => "12.0" }
   s.source       = { :git => "https://github.com/kirillzyusko/react-native-keyboard-controller.git", :tag => "#{s.version}" }
 
   s.source_files = "ios/**/*.{h,m,mm,swift}"


kirillzyusko:
@matinzd hm, interesting! I'm using RN 0.75.2 and in Fabric example I haven't seen such errors 🤔 

Do you experience similar issues with other libraries, such as `react-native-screens`/`react-native-gesture-handler`? They are also using minimal supported `11` iOS version 🤔 

Can you maybe share full stack trace? Where this `.value` come from? On top of my head I've seen this when I had Flipper enabled...

matinzd:
It's weird. I cannot find it anymore after increasing the minimum deployment. I will go back to 11 to see if I can reproduce it.

t4ngth00:
I got the same issue with RN version 0.75.3

kirillzyusko:
@t4ngth00 do you use flipper? 👀

What I'm currently trying to understand os why it fails if I don't use any platform dependent code? I have some conditional compilable code (like use new API for iOS 13 and use stub for older versions), but can such checks really lead to compilation errors? 🤔

@t4ngth00 would you mind to share a stacktrace of the compilation error? I.e. which exactly file triggers the error?

t4ngth00:
> do you use flipper?

No I don't



> would you mind to share a stacktrace of the compilation error? I.e. which exactly file triggers the error?

I'm sorry that I have just gone for vacation and will be back next week. But you can simply reproduce the issue by upgrade react native 0.75.3 and compile with Xcode.

kirillzyusko:
@t4ngth00 I updated RN to 0.75.3 in example app but I haven't seen mentioned problems 🤔 PR for a reference: https://github.com/kirillzyusko/react-native-keyboard-controller/pull/586

matinzd:
> @t4ngth00 I updated RN to 0.75.3 in example app but I haven't seen mentioned problems 🤔 PR for a reference: #586

What xcode are you using?

matinzd:
Very weird I cannot reproduce it anymore. Can you try it in a clean environment with new arch turned on?

kirillzyusko:
> Which version of xcode are you using?

Version 15.4 (15F31d). Which version do you use?

> Very weird I cannot reproduce it anymore. Can you try it in a clean environment with new arch turned on?

Yeah, I'll try to create a new example app and see if I can reproduce it there 👀

matinzd:
> Version 15.4 (15F31d). Which version do you use?

Same here :D (15F31d)

matinzd:
I am gonna close this. It was a very random thing. It seems like the min deployment was not being properly applied in pod install.

kirillzyusko:
Okay @matinzd 

Thank you for the contribution anyway. If you discover this problem again and it'll be always reproducible feel free to open a new/re-open this PR and provide some more details 😊

PR #603, opened 2024-09-27
Make KeyboardToolbar inherit ViewProps & export default toolbar height

### LINKED ISSUE #601: Know the KeyboardToolbar height
Hey 👋 first of all thanks for this lib, it's solving a lot of problems.

**Is your feature request related to a problem? Please describe.**
I'm using the KeyboardToolbar component and it works great. I often find myself wanting to space the input with the keyboard while taking into account the toolbar height. I know it's 42 because i measured it so I've included something like`bottomSpace = actualOffset + TOOLBAR_HEIGHT` and use it in my component. 

**Describe the solution you'd like**
I feel like this should be provided by the lib though. One way of doing it would be to allow to forward props to the root view. This way we can write
 ```Typescript <KeyboardToolbar onLayout={handleLayout} />```
We can retrieve the height directly in the onLayout event. What do you think of this ?
Do you have a better idea ? I can definitely make the implementation changes if it's ok for you.

**Describe alternatives you've considered**
None.

**Additional context**

CorentinAndre:
## 📜 Description

2 improvements:
- make the KeyboardToolbar inherit ViewProps to support complex cases where you want to know the height of the KeyboardToolbar with an onLayout event
- exports the KEYBOARD_TOOLBAR_HEIGHT to easily add the correct offset on KeyboardAwareScrollView component

## 💡 Motivation and Context

The library did not provide any way of knowing the height of the toolbar. I've done both solutions because I find it easier to just grab the constant to set the bottom offset @kirillzyusko. Let me know if it's an issue :)

I don't know if I should add tests for this kind of stuff, please let me know.

Fixes #601.

### JS

- Exports `KEYBOARD_TOOLBAR_HEIGHT` constant
- make `KeyboardToolbar` inherit `ViewProps` from `react-native`

Patched the lib locally and verified it works.

## 📝 Checklist

- [ ] CI successfully passed


src/components/KeyboardToolbar/index.tsx
```diff
@@ -146,7 +150,7 @@ const KeyboardToolbar: React.FC<KeyboardToolbarProps> = ({
 
   return (
     <KeyboardStickyView offset={offset}>
-      <View style={toolbarStyle} testID={TEST_ID_KEYBOARD_TOOLBAR}>
+      <View style={toolbarStyle} testID={TEST_ID_KEYBOARD_TOOLBAR} {...rest}>
```
CorentinAndre, line 153: didn't work if I spreaded the props to the KeyboardStickyView. Since the underlying component is a Reanimated view, I suspect it kind of messes with onLayout events

src/components/KeyboardToolbar/index.tsx
```diff
@@ -57,7 +60,7 @@ const TEST_ID_KEYBOARD_TOOLBAR_NEXT = `${TEST_ID_KEYBOARD_TOOLBAR}.next`;
 const TEST_ID_KEYBOARD_TOOLBAR_CONTENT = `${TEST_ID_KEYBOARD_TOOLBAR}.content`;
 const TEST_ID_KEYBOARD_TOOLBAR_DONE = `${TEST_ID_KEYBOARD_TOOLBAR}.done`;
 
-const KEYBOARD_TOOLBAR_HEIGHT = 42;
+export const KEYBOARD_TOOLBAR_HEIGHT = 42;
```
kirillzyusko, line 63: Are we sure we want to export that? 😅 

I created it as an internal variable and didn't suppose it should be exported. The thing that I'm afraid - if in future the size will be dynamic (based on accessibility preferences, for example), then the height will not be a constant anymore, so I think that deriving any layout information via `onLayout` callback is a more preferrable approach at the moment.

What do you think? 👀

src/components/KeyboardToolbar/index.tsx
kirillzyusko: Yeah @CorentinAndre 

In my experiments it always reported `height` as `0` 🤷‍♂️

src/components/KeyboardToolbar/index.tsx
CorentinAndre: Yup it might be creating a breaking change for something that can be retrieved by the onLayout. I'll remove the commit

CorentinAndre: Yes, I wasn't sure either. I'll remove the commit

docs/docs/api/components/keyboard-toolbar/index.mdx
```diff
@@ -280,6 +281,8 @@ import { StyleSheet, Text, TextInput as TextInputRN, View } from "react-native";
 import {
   KeyboardAwareScrollView,
   KeyboardToolbar,
+  // This is useful when you want to offset your content based on the toolbar default height.
+  KEYBOARD_TOOLBAR_HEIGHT,
```
kirillzyusko: @CorentinAndre Please, revert this too 🙏

docs/docs/api/components/keyboard-toolbar/index.mdx
```diff
@@ -36,6 +36,7 @@ import toolbar from "./toolbar.lottie.json";
 - **Supports dark/light theme** 🌓: Adapts to the theme settings of the user's device for a seamless experience.
 - **Extended accessibility support** 🔍: Ensures that all users, including those with disabilities, can navigate through inputs effectively.
 - **Full control over the buttons behavior** 🔧: Customize the actions triggered by the next, previous, and done buttons according to your needs.
+- **Extends ViewProps** 📜: Supports all the props that `View` component has.
 
 ## Props
```
kirillzyusko, line 42: Can we add it here something like:

```
### View Props

Inherits [View Props](https://reactnative.dev/docs/view#props).
```

RN website uses such style, for example in: https://reactnative.dev/docs/keyboardavoidingview#view-props

kirillzyusko (approved):
Looks good - just left two minor improvements!

Approving PR to run CI and see if it's alright or not 👀

docs/docs/api/components/keyboard-toolbar/index.mdx
kirillzyusko: For transparency: I'd like to keep `**Extends ViewProps** 📜: Supports all the props that `View` component has.` text, but also would like to add a link to ViewProps (in the same way as official RN docs do)

kirillzyusko:
Two jobs failed - this is expectable, because you are not a person who owes a repository. I really would like to fix the CI setup, but don't hve time for that 🙈

kirillzyusko:
Thank you for your contribution @CorentinAndre ❤️

PR #621, opened 2024-10-08
fix: make `getTopWindow` more robust

### LINKED ISSUE #618: [iOS] `OverKeyboardView` does not work when the keyboard is not active
**Describe the bug**

When I use this component, it works as expected on Android, but on iOS, the children of `OverKeyboardView` are invisible when the keyboard is not active. I know they are rendered because they show up in the element inspector and can be interacted with (buttons are still pressable).

I think this happens because `UIRemoteKeyboardWindow` is not available when the keyboard is not active, but I don't think it is expected for the contents to be invisible in such cases. If you look at Telegram, the popover can be opened whether or not the keyboard is active.

**Code snippet**

```tsx
<OverKeyboardView visible={popoverModalVisible}>
  <View
    className={cn('absolute inset-0', className)}
    accessibilityViewIsModal={popoverModalVisible}
    onAccessibilityEscape={onRequestClose}
    pointerEvents="box-none"
    style={style}>
    {popoverContent}
  </View>
</OverKeyboardView>
```

**Repo for reproducing**

Please let me know if you need a repro, but I believe it should be fairly easy to reproduce.

**To Reproduce**

I tried my best to run the example ios app but I'm getting this error:

Screenshot 2024-10-08 at 12 07 57 AM

**Expected behavior**

When the keyboard is active, the children in `OverKeyboardView` should not be invisible.

**Screenshots**

### iOS

simulator_screenshot_ABDF4AEB-BC2F-4088-AB87-DD494B823B18

### Android

Screenshot_1728317773

- Device: iPad mini
- OS: iOS 18.1
- RN version: 0.75.3
- RN architecture: Paper
- JS engine: Hermes
- Library version: 1.14.0

thespacemanatee:
## 📜 Description

This PR updates the top window retrieval method:

Old: `UIApplication.shared.windows.last` (deprecated in iOS 15+)
New: `UIWindowScene`-based approach

Differences:

- `UIApplication.shared.windows.last`:
    - It accesses all windows associated with the application.
    - The last window in this array isn't always the topmost or key window, especially in more complex app setups.
    - This approach is deprecated as of iOS 15, though it still works in many cases.
- `UIWindowScene`:
    - Iterates through all connected scenes to find the active foreground scene
    - Ensures selection of the correct key window in multi-window environments (e.g., Split View in iPadOS)
    - Provides more reliable behaviour in complex app configurations that utilize multiple `UIWindowScenes`

## 💡 Motivation and Context

Closes #618

In our case, this change was required to get the children of the `OverKeyboardView` to show. Without it, they were still rendered, but invisible. This change should enhance the reliability across different iOS versions and device types.

### iOS

- Get the last key window in the current `UIWindowScene` as opposed to the last window of all windows associated with the app

It works on my company project. Did not manage to get the example app to compile on my machine. Could be due to the (lack of) provisioning profiles.

## 📝 Checklist

- [x] CI successfully passed

### DIFF
--- a/ios/extensions/UIApplication.swift
+++ b/ios/extensions/UIApplication.swift
@@ -9,6 +9,22 @@ import Foundation
 import UIKit
 
 public extension UIApplication {
+  var activeWindow: UIWindow? {
+    if #available(iOS 13.0, *) {
+      for scene in connectedScenes {
+        if scene.activationState == .foregroundActive,
+           let windowScene = scene as? UIWindowScene,
+           let keyWindow = windowScene.windows.first(where: { $0.isKeyWindow })
+        {
+          return keyWindow
+        }
+      }
+      return nil
+    } else {
+      return windows.last { $0.isKeyWindow }
+    }
+  }
+
   static func topViewController(
     base: UIViewController? = UIApplication.shared.keyWindow?.rootViewController
   ) -> UIViewController? {
--- a/ios/extensions/UIWindow.swift
+++ b/ios/extensions/UIWindow.swift
@@ -38,7 +38,7 @@ public extension UIWindow {
 
     func getTopWindow() -> UIWindow? {
       // Return the keyboard window if it's available, otherwise return the last window
-      return keyboardWindow ?? UIApplication.shared.windows.last
+      return keyboardWindow ?? UIApplication.shared.activeWindow
     }
   }
 


ios/extensions/UIWindow.swift
```diff
@@ -38,7 +38,22 @@ public extension UIWindow {
...
+          if scene.activationState == .foregroundActive,
+            let windowScene = scene as? UIWindowScene,
+            let keyWindow = windowScene.windows.first(where: { $0.isKeyWindow }) {
+            return keyWindow
+          }
+        }
+        return nil
+      } else {
+        return UIApplication.shared.windows.last { $0.isKeyWindow }
+      }
```
kirillzyusko, line 56: The changes looks good to me, but may I ask you to slightly re-work this part?

I would like to have this code somewhere isolated, because potentially we may want to get an access to main window in future, so I would move this code to an extension:

```swift
extension UIApplication {
    var activeWindow: UIWindow? {
        if #available(iOS 13.0, *) {
            for scene in connectedScenes {
                if scene.activationState == .foregroundActive,
                   let windowScene = scene as? UIWindowScene,
                   let keyWindow = windowScene.windows.first(where: { $0.isKeyWindow }) {
                    return keyWindow
                }
            }
            return nil
        } else {
            return windows.last { $0.isKeyWindow }
        }
    }
}
```

And then use it like:

```swift
return keyboardWindow ?? UIApplication.shared.activeWindow
```

Not sure if this code works - just wanted to show the idea. Hope you can test these modifications in your project 👀

kirillzyusko:
And yeah, if possible, please fix `swiftlint` and `swiftformat` jobs 👀 

I think running `swiftformat` should be enough and `swiftlint` should be fixed automatically

ios/extensions/UIWindow.swift
thespacemanatee: Will try and let you know!

ios/extensions/UIApplication.swift
```diff
@@ -9,6 +9,22 @@ import Foundation
...
+        }
+      }
+      return nil
+    } else {
+      return windows.last { $0.isKeyWindow }
+    }
+  }
+
   static func topViewController(
     base: UIViewController? = UIApplication.shared.keyWindow?.rootViewController
```
thespacemanatee, line 29: @kirillzyusko Xcode also complains about using `UIApplication.shared.keyWindow`, I haven't tested it yet but replacing this with the `activeWindow` extension removes the warning and should achieve the same thing.

Screenshot 2024-10-09 at 2 19 47 AM

ios/extensions/UIWindow.swift
thespacemanatee: @kirillzyusko Looks like it functionally works the same, but I found a new issue that I'm not sure if is the result of this change, or also present in the existing code.

In my app, I have routes guarded behind an auth check, and the `Screens` rendered in the `Navigator` change based on the auth state. When I open the `OverKeyboardView` for the first time, it works, and navigation events such as `push` can be done and `OverKeyboardView` still works. But if the user logs out, the `Screens` are remounted and the next time I try to open this `OverKeyboardView`, the children become invisible again.

Do you think this PR should be put on hold until I figure out the fix for that as well?

ios/extensions/UIApplication.swift
kirillzyusko: Yes @thespacemanatee I'm also constantly getting these warnings 🙈 

I would split changes in two parts anyway 🙃 Like merge these changes and then open a second PR that will repace a single line.

In this case it'll be easier to manage 😊

kirillzyusko (approved):

kirillzyusko:
Merged @thespacemanatee 🙌 

Would you mind to submit a PR that we discussed [here](https://github.com/kirillzyusko/react-native-keyboard-controller/pull/621#discussion_r1792319543)?

ios/extensions/UIWindow.swift
kirillzyusko: Ah, sad to hear that it still has issues 😔 

Would you mind to submit a new issues and describe hte problem? Just want to track it somewhere so that I know that I'll not forget about the problem and will return to it :)

thespacemanatee: @kirillzyusko let me create an issue when I get some time this week! Thanks man

PR #627, opened 2024-10-10
fix: broken `StatusBar` when `Activity` changes

war-in:
## 📜 Description

I suggest keeping track of the last activity and if it changes we should update the controller with the current window

## 💡 Motivation and Context

StatusBar is not responding when the activity changes - see [issue](https://github.com/Expensify/App/issues/48782#issue-2513801944).

The problem is located [here](https://github.com/kirillzyusko/react-native-keyboard-controller/blob/bb5e9e1d1e00278ce800b9558b75e56ff799abda/android/src/main/java/com/reactnativekeyboardcontroller/modules/StatusBarManagerCompatModuleImpl.kt#L74) in `getController` method.

We check if the controller is not `null` but if the activity changes (is closed and reopened) the controller is initialized and points to a non-existent window (taken from the previous activity), so we won't update the controller.

Outdated controller can't control the StatusBar (it doesn't have access to the current window)

### Android

- fixed status bar when activity changes

Tested manually in Expensify project.

### Android only

I tested it with Expensify HybridApp. For those who don't have access to it, the flow looks like this:
- open activity A (which doesn't use keyboard controller)
- activity A opens activity B (which is New Expensify app)
- activity B can control the status bar
- close activity B
- reopen activity B
- the status bar can't be modified

|Before|After|
|-------|-----|
||

## 📝 Checklist

- [x] CI successfully passed

### DIFF
--- a/android/src/main/java/com/reactnativekeyboardcontroller/modules/StatusBarManagerCompatModuleImpl.kt
+++ b/android/src/main/java/com/reactnativekeyboardcontroller/modules/StatusBarManagerCompatModuleImpl.kt
@@ -2,6 +2,7 @@ package com.reactnativekeyboardcontroller.modules
 
 import android.animation.ArgbEvaluator
 import android.animation.ValueAnimator
+import android.app.Activity
 import android.os.Build
 import androidx.annotation.RequiresApi
 import androidx.core.view.WindowInsetsCompat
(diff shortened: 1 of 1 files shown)


kirillzyusko:
The code looks good ❤️ Thank you for the contribution! 🚀

Issue #259, opened 2023-10-20
Proposal: Merge with react-native-bars project


zoontek:
Hi 👋

First, thanks for your work on this library, this is very much needed in the ecosystem!

I'm currently maintaining [`react-native-bars`](https://github.com/zoontek/react-native-bars), a library for fully transparent (and fallback to translucent) system bars. Compared to `KeyboardProvider` `statusBarTranslucent` and `navigationBarTranslucent`, it provides an android theme that avoid the initial blink (before the React app kicks in) + provides some useful components to handle `barStyle` (similarly as the builtin `StatusBar` component).

As keyboard handling is super related to system bars, I think it could be great to merge those libs? (I even think that it should enforce fully transparent UI without any choice, exactly like `react-native-bars` or iOS + recommend `react-native-safe-area-context` for dealing with safe areas in a similar way on both OS, but it's my personal opinion 😅)

kirillzyusko:
Hello @zoontek 👋 

I understand the motivation to merge these projects (because of native code in Android).

But I'm just thinking, that people who wants to control `NavigationBar` may not understand why do they need to install library that helps you to deal with keyboard? 🤔 What do you think about it?

kirillzyusko:
Yes @zoontek it makes sense what you are saying! I just was confused about SEO search (obviously name `react-native-keyboard-controller` doesn't give a confidence that it's a right library for managing Navigation and System bar 😅).

In order to merge projects we need:
- to have `init` method for overriding a theme to avoid a blink;
- new NativeModule for managing NavigationBar and StatusBar properties;
- components in JS (NavigationBar, StatusBar).

Did I miss anything?

zoontek:
@kirillzyusko I think there's no need for the `init` method in `react-native-keyboard-controller`, as setting a parent style (`Theme.EdgeToEdge) + waiting for `KeyboardProvider` to kicks in should do the job.

For the rest, that's it, yeah. Maybe I could experiment on a branch if you are OK with that?

kirillzyusko:
@zoontek yes, sure, feel free to experiment with it 👍  😊

zoontek:
I noticed something: `navigationBarTranslucent` currently doesn't seem to enable [edge-to-edge layout](https://developer.android.com/develop/ui/views/layout/edge-to-edge-manually).
Navigation bar stays dark and the UI is not drawn under:

[...]

PR #660, opened 2024-10-27
Detect react-native-edge-to-edge and bypass statusBarTranslucent / navigationBarTranslucent

zoontek:
## 📜 Description

This PR brings `react-native-edge-to-edge` detection to `react-native-keyboard-controller`.

## 💡 Motivation and Context

The future of Android is [edge-to-edge](https://developer.android.com/about/versions/15/behavior-changes-15#edge-to-edge), and to make the React Native developer experience seamless in this regard, the ecosystem needs to transition from “opaque system bars by default” to “edge-to-edge by default.”

To prevent library authors from implementing their own edge-to-edge solutions—which could interfere with other libraries—and because it’s not possible to reliably detect if edge-to-edge is already enabled on Android, we have collaborated with [Expo](https://expo.dev/) to create a library that handles this functionality and is detectable using a simple helper: [`react-native-is-edge-to-edge`](https://github.com/zoontek/react-native-edge-to-edge/tree/main/react-native-is-edge-to-edge).

This approach allows you to bypass certain options and props (in this case, `KeyboardProvider` `statusBarTranslucent` and `navigationBarTranslucent`), helping to avoid frustrating issues like, “My app starts in edge-to-edge, but the system bars reappear immediately.” (because RNKC causes them to reappear)

This PR is intentionally lightweight, as RNKC doesn’t seem to interfere with edge-to-edge when `statusBarTranslucent` and `navigationBarTranslucent` are set to `true`. However, if necessary, you could also use a small helper in Kotlin to skip part of your library code if needed:

```kt
object EdgeToEdge {
  val ENABLED: Boolean
    get() = try {
      // we cannot detect edge-to-edge, but we can detect react-native-edge-to-edge install
      Class.forName("com.zoontek.rnedgetoedge.EdgeToEdgePackage")
      true
    } catch (exception: ClassNotFoundException) {
      false
    }
}
```

Closes https://github.com/kirillzyusko/react-native-keyboard-controller/issues/259

## 📢 Changelog

- Detect `react-native-edge-to-edge` to potentially bypass `statusBarTranslucent` / `navigationBarTranslucent` values.

### JS

- Added a warning about `statusBarTranslucent` / `navigationBarTranslucent` usage when `react-native-edge-to-edge` is installed.

### iOS

- No change (native codebase) ❌

### Android

- No change (native codebase) ❌ 

Install [`react-native-edge-to-edge`](https://github.com/zoontek/react-native-edge-to-edge) in the example app.

Screenshot 2024-10-26 at 12 49 14

## 📝 Checklist

- [x] CI successfully passed

### DIFF
--- a/example/yarn.lock
+++ b/example/yarn.lock
@@ -4747,10 +4747,10 @@ react-native-haptic-feedback@2.3.1:
   resolved "https://registry.yarnpkg.com/react-native-haptic-feedback/-/react-native-haptic-feedback-2.3.1.tgz#2ef4ac7d4f63ac06bd64b659f509362f127b16e5"
   integrity sha512-dPfjV4iVHfhVyfG+nRd88ygjahbdup7KFZDM5L2aNIAzqbNtKxHZn5O1pHegwSj1t15VJliu0GyTX7XpBDeXUw==
 
-react-native-is-edge-to-edge@^1.1.0:
-  version "1.1.0"
-  resolved "https://registry.yarnpkg.com/react-native-is-edge-to-edge/-/react-native-is-edge-to-edge-1.1.0.tgz#45c0fb4ad2971b6c8dec49f4ace29bea4cf01b40"
-  integrity sha512-WQ14EZDVchiFQ8Xs1KtPhkSPFKHdIm/BVQoElyy6SjzDfFlhP0ERFkAAPmdgmmHjWwYyjm+ibvq1PieT8Xfpbw==
+react-native-is-edge-to-edge@^1.1.1:
+  version "1.1.1"
+  resolved "https://registry.yarnpkg.com/react-native-is-edge-to-edge/-/react-native-is-edge-to-edge-1.1.1.tgz#78ee045dab74868a4d1c37df9058c4bdecc28e62"
+  integrity sha512-F/CPckyjLgQNOv8BaUorSyjNv7Ev3eLuzWT+BrIp4504Zw1LAa44M3FPGz5E2nZMcvHHbSmBLHei5M3g+gmLHA==
 
 "react-native-keyboard-controller@link:..":
   version "0.0.0"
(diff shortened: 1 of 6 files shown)


src/animated.tsx
```diff
@@ -189,13 +190,27 @@ export const KeyboardProvider = ({
...
+  }, [statusBarTranslucent, navigationBarTranslucent]);
+
   return (
     <KeyboardContext.Provider value={context}>
       <KeyboardControllerViewAnimated
         ref={viewTagRef}
         enabled={enabled}
-        navigationBarTranslucent={navigationBarTranslucent}
-        statusBarTranslucent={statusBarTranslucent}
+        navigationBarTranslucent={isEdgeToEdge() || navigationBarTranslucent}
```
kirillzyusko, line 212: A small hint - let's maybe create a global variable:

```ts
const IS_EDGE_TO_EDGE = isEdgeToEdge()
```

And re-use it? Just thinking, that it can be calculated one time and re-use result. It'll be more optimal, since we will not call a function (and perform comparison) two times (or 3 times in debug).

What do you think?

zoontek: It returns a pre-computed constant, so the impect is just a function call, but I agree, for clarity it's better to update it.

kirillzyusko (approved):

example/yarn.lock
```diff
@@ -4747,6 +4747,11 @@ react-native-haptic-feedback@2.3.1:
   resolved "https://registry.yarnpkg.com/react-native-haptic-feedback/-/react-native-haptic-feedback-2.3.1.tgz#2ef4ac7d4f63ac06bd64b659f509362f127b16e5"
   integrity sha512-dPfjV4iVHfhVyfG+nRd88ygjahbdup7KFZDM5L2aNIAzqbNtKxHZn5O1pHegwSj1t15VJliu0GyTX7XpBDeXUw==
 
+react-native-is-edge-to-edge@^1.0.0:
```
kirillzyusko, line 4750: Another point - you should also run `yarn` command in `FabricExample` to update `yarn.lock` there as well 👀

Issue #649, opened 2024-10-22
Reanimated warning regarding `value` render during component render.


akuul:
**Describe the bug**
Since `React Native Reanimated v3.16.0` changed it's logger, I'm getting 
```bash
[Reanimated] Reading from `value` during component render. Please ensure that you do not access the value` property or use `get` method of a shared value while React is rendering a component.
```
when using `KeyboardAwareView` component. Currently there are few packages having this problem with open issues like `@gorhom/react-native-bottom-sheet` or `react-native-pager-view`.

This seems to be related to `KeyboardAvoidingView/hooks.ts` where it uses context to get `reanimated` value from `useKeyboardContext()` hook. 

Very quick playaround with the snippet below fixes the warning, but this doesn't look like ideal solution.

```typescript
  const heightWhenOpened = useSharedValue(0);
  const height = useSharedValue(0);
  const progress = useSharedValue(0);
  const isClosed = useSharedValue(true);

  useEffect(() => {
    heightWhenOpened.value = -reanimated.height.value;
    height.value = -reanimated.height.value;
    progress.value = reanimated.progress.value;
    isClosed.value = reanimated.progress.value === 0;
  }, []);
  ```

**Code snippet**
```typescript
function App() {
  return (
    <KeyboardProvider>
      <KeyboardAvoidingView></KeyboardAvoidingView>
    </KeyboardProvider>
  );
}

export default App;

```
**Expected behavior**
React Native Reanimated should not throw an error

**Screenshots**
![image](https://github.com/user-attachments/assets/36bc9061-0f52-4b52-b664-c956993eda58)

**Smartphone (please complete the following information):**

- RN version: 0.75.4
- RN architecture: Old
- JS engine: Hermes
- Library version: 1.14.2

**Additional context**
This issue does not appear with `KeyboardAwareScrollView`.

kirillzyusko:
Okay, interesting, thank you @akuul 🙌 

I'll have a look on that 👀 

> This seems to be related to KeyboardAvoidingView/hooks.ts where it uses context to get reanimated value from useKeyboardContext() hook.

You are right, this code triggers that. But I do not understand this restriction, because I know that only for `ref` variables it's not recommended to read/mutate `.current` value. Regarding `useSharedValue` - it's based on `useState`, so I don't understand yet why it's forbidden to read `.value` yet.

But thank you for opening the issue ❤️ I'll figure out how to re-work the code to fix it!

[...]

PR #662, opened 2024-10-27
fix: avoid shared value reads during render

kirillzyusko:
## 📜 Description

Fixed popping up warnings when re-render happens if `KeyboardAvoidingView` is used.

## 💡 Motivation and Context

Reading `.value` during render is a bad approach because it violates rules of react, so we shouldn't use it anymore.

Fortunately we have only one place in the codebase - it's `KeyboardAvoidingView`. Technically I don't violate rules of 

Initially I had 3 approaches in my head:

### 1️⃣  Precompute initial value via `useMemo`

The first approach I was thinking of was this:

```tsx
const {height, progress} = useMemo(() => ({
  height: context.reanimated.height,
  progress: context.reanimated.progress,
}), []);

const heightWhenOpened = useSharedValue(-height);
const height = useSharedValue(-height);
const progress = useSharedValue(progress);
const isClosed = useSharedValue(progress === 0);
```

However later on in react docs I found this:

image

So I thought that in future **potentially** `useMemo` may be re-evaluated, so we'll get this warning again. Which is not desirable.

### 2️⃣ Memoize the initial value via `useState`

Another approach was the usage of `useState`:

```tsx
  const [initialHeight] = useState(() => -reanimated.height.value);
  const [initialProgress] = useState(() => reanimated.progress.value);

  const heightWhenOpened = useSharedValue(initialHeight);
  const height = useSharedValue(initialHeight);
  const progress = useSharedValue(initialProgress);
  const isClosed = useSharedValue(initialProgress === 0);
```

In this case we derive value (kind of preparing them) and we are sure that those values will not be accidentally re-calculated (because in such way state will be initialized only one time).

The only downside is that we are creating additional variables and we use state not for its purpose (such values will never be updated actually).

### 3️⃣ create `useLazySharedValue` hook

The implementation could look like:

```tsx
export function useLazySharedValue<Value>(initialValue: () => Value): SharedValue<Value> { // function instead of value
  const [mutable] = useState(() => makeMutable(initialValue()));
  useEffect(() => {
    return () => {
      cancelAnimation(mutable);
    };
  }, [mutable]);
  return mutable;
}

// and usage...

const progress = useLazySharedValue(() => context.reanimated.progress);
```

However I have next concerns in my head:
- `makeMutable` is publicly exported, but I don't want to copy internal implementation of `useSharedValue` (even though this hook is a very simple). If it changes over time I'll have to replicate these changes in this package as well.

<hr>

Keeping all possible ways in my head I decided to go with approach **2**. In future I can adjust the logic if needed.

Closes https://github.com/kirillzyusko/react-native-keyboard-controller/issues/649

### JS

- prepare init values from shred value via `useState(() => `;

Tested locally on iPhone 16 Pro iOS 18.0.

|Before|After|
|-------|-----|
|<video src="https://github.com/user-attachments/assets/6fe02687-683a-4988-a370-3f1fba864c04">|<video src="https://github.com/user-attachments/assets/c7a79568-8705-49f8-97fd-19540291b2a9">|

## 📝 Checklist

- [x] CI successfully passed

### DIFF
--- a/src/components/KeyboardAvoidingView/hooks.ts
+++ b/src/components/KeyboardAvoidingView/hooks.ts
@@ -1,3 +1,4 @@
+import { useState } from "react";
 import { useSharedValue } from "react-native-reanimated";
 
 import { useKeyboardContext } from "../../context";
@@ -5,10 +6,15 @@ import { useKeyboardHandler } from "../../hooks";
 
 export const useKeyboardAnimation = () => {
   const { reanimated } = useKeyboardContext();
-  const heightWhenOpened = useSharedValue(-reanimated.height.value);
-  const height = useSharedValue(-reanimated.height.value);
-  const progress = useSharedValue(reanimated.progress.value);
-  const isClosed = useSharedValue(reanimated.progress.value === 0);
+
+  // calculate it only once on mount, to avoid `SharedValue` reads during a render
+  const [initialHeight] = useState(() => -reanimated.height.value);
+  const [initialProgress] = useState(() => reanimated.progress.value);
+
+  const heightWhenOpened = useSharedValue(initialHeight);
+  const height = useSharedValue(initialHeight);
+  const progress = useSharedValue(initialProgress);
+  const isClosed = useSharedValue(initialProgress === 0);
 
   useKeyboardHandler(
     {


kirillzyusko:
@IvanIhnatsiuk would you mind to have a look?

src/components/KeyboardAvoidingView/hooks.ts
```diff
@@ -1,14 +1,20 @@
...
 export const useKeyboardAnimation = () => {
   const { reanimated } = useKeyboardContext();
-  const heightWhenOpened = useSharedValue(-reanimated.height.value);
-  const height = useSharedValue(-reanimated.height.value);
-  const progress = useSharedValue(reanimated.progress.value);
-  const isClosed = useSharedValue(reanimated.progress.value === 0);
+
+  // calculate it only once on mount, to avoid `SharedValue` reads during a render
+  const [initialHeight] = useState(() => -reanimated.height.value);
+  const [initialProgress] = useState(() => reanimated.progress.value);
```
IvanIhnatsiuk, line 12: hm, they already use useState under the hood https://github.com/software-mansion/react-native-reanimated/blob/main/packages/react-native-reanimated/src/hook/useSharedValue.ts#L19 and if we have a look at the react docs 

initialState: The value you want the state to be initially. It can be a value of any type, but there is a special behavior for functions. This argument is ignored after the initial render.

So, are you sure that you need these changes?

kirillzyusko: Yeah, but new REA 3.16 will complain if you use `.value` during render. So you should void the usage of `.value` during re-render (but you can use for the first render).

I had several approaches:

1. To use `useMemo` to calculate initial values for `reanimated.*.value` - but this is not reliable, because `useMemo` in future may throw cache away and re-calculate memoization.

2. To use `useState` to calculate initial value.

3. to create own `useLazySharedValue` hook with next implementation:

```ts
export function useSharedValue<Value>(initialValue: () => Value): SharedValue<Value> { // function instead of value
  const [mutable] = useState(() => makeMutable(initialValue()));
  useEffect(() => {
    return () => {
      cancelAnimation(mutable);
    };
  }, [mutable]);
  return mutable;
}
```

If you have more ideas @IvanIhnatsiuk then I'll be happy to consider your thoughts 😊

IvanIhnatsiuk: hm, looks like it would be better to implement your own hook and remove it once people migrate to REA 3.16

Also, it helps to remove code duplication like this:

```.tsx
 const [initialHeight] = useState(() => -reanimated.height.value);
 const heightWhenOpened = useSharedValue(initialHeight);
```

and replace it with your own hook

```.tsx
 const heightWhenOpened = useLazySharedValue(() => -reanimated.height.value);
```

But anyway, it's up to you 🙂

kirillzyusko: Well, there are few reasons why I don't want to implement my hook:
- `makeMutable` is publicly exported, but I don't want to copy internal implementation of `useSharedValue` (even though this hook is a very simple). If it changes over time I'll have to replicate these changes in this package as well.
- we have kind of code duplication, but thus we extract values and then reuse the value (I'm not sure if reading `SharedValue` on first render is an expensive operation or not, but I saw somewhere that it adds an overhead - but maybe mistaken).

This is why I think `useState` approach is better. anyway, it's pretty minor code modification - in future we may revisit this code 🙂

PR #668, opened 2024-10-30
fix: inverted list padding

lpatrun:
## 📜 Description

I noticed the padding is incorrectly added if the Flatlist/Flashlist is inverted. Also, I have noticed that because of the bottom tab navigation, the offset when the keyboard is opened is not correctly calculated and the bottom tab height should be deducted.

## 💡 Motivation and Context

This problem fixes padding on inverted flatlist/flashlist if the keyboard is opened.

### JS

- added inverted handling
- added property to be able to adjust the offset when the keyboard is opened

It was tested in the app.

### DIFF
--- a/src/components/KeyboardAwareScrollView/index.tsx
+++ b/src/components/KeyboardAwareScrollView/index.tsx
@@ -38,6 +38,10 @@ export type KeyboardAwareScrollViewProps = {
   enabled?: boolean;
   /** Adjusting the bottom spacing of KeyboardAwareScrollView. Default is `0` */
   extraKeyboardSpace?: number;
+   /** Used if inverted Flatlist/Flashlist is being used */
+  inverted?: boolean;
+  /** Ajdusting the offset for the content when keyboard is shown on inverted Flatlist/Flashlist on screen with BottomTab navigation */
+  tabBarHeight?: number;
   /** Custom component for `ScrollView`. Default is `ScrollView` */
   ScrollViewComponent?: React.ComponentType<ScrollViewProps>;
 } & ScrollViewProps;
@@ -94,6 +98,8 @@ const KeyboardAwareScrollView = forwardRef<
       extraKeyboardSpace = 0,
       ScrollViewComponent = Reanimated.ScrollView,
       snapToOffsets,
+      inverted = false,
+      tabBarHeight = 0,
       ...rest
     },
     ref,
@@ -349,8 +355,9 @@ const KeyboardAwareScrollView = forwardRef<
     );
 
     const view = useAnimatedStyle(
-      () =>
-        enabled
+      () => {
+        const invertedOffset = inverted ? -tabBarHeight : 0;
+        return enabled
           ? {
               // animations become choppy when scrolling to the end of the `ScrollView` (when the last input is focused)
               // this happens because the layout recalculates on every frame. To avoid this we slightly increase padding
@@ -358,9 +365,10 @@ const KeyboardAwareScrollView = forwardRef<
               // from 0 to `keyboardHeight`, and here our padding is `keyboardHeight + 1`. It allows us not to re-run layout
               // re-calculation on every animation frame and it helps to achieve smooth animation.
               // see: https://github.com/kirillzyusko/react-native-keyboard-controller/pull/342
-              paddingBottom: currentKeyboardFrameHeight.value + 1,
+              paddingBottom: currentKeyboardFrameHeight.value + invertedOffset + 1,
             }
-          : {},
+            : {}
+        },
       [enabled],
     );
 
@@ -371,8 +379,9 @@ const KeyboardAwareScrollView = forwardRef<
         scrollEventThrottle={16}
         onLayout={onScrollViewLayout}
       >
+        {inverted ? <Reanimated.View style={view} /> : null}
         {children}
-        <Reanimated.View style={view} />
+        {!inverted ? <Reanimated.View style={view} /> : null}
       </ScrollViewComponent>
     );
   },


src/components/KeyboardAwareScrollView/index.tsx
```diff
@@ -38,6 +38,10 @@ export type KeyboardAwareScrollViewProps = {
   enabled?: boolean;
   /** Adjusting the bottom spacing of KeyboardAwareScrollView. Default is `0` */
   extraKeyboardSpace?: number;
+   /** Used if inverted Flatlist/Flashlist is being used */
+  inverted?: boolean;
+  /** Ajdusting the offset for the content when keyboard is shown on inverted Flatlist/Flashlist on screen with BottomTab navigation */
+  tabBarHeight?: number;
```
kirillzyusko, line 44: I think you can re-use `extraKeyboardSpace` property for these purposes? From docs:

image

lpatrun:
@kirillzyusko I will try your suggestion and get back to you. Thank you for your promptness

src/components/KeyboardAwareScrollView/index.tsx
```diff
@@ -38,6 +38,10 @@ export type KeyboardAwareScrollViewProps = {
   enabled?: boolean;
   /** Adjusting the bottom spacing of KeyboardAwareScrollView. Default is `0` */
   extraKeyboardSpace?: number;
+   /** Used if inverted Flatlist/Flashlist is being used */
+  inverted?: boolean;
```
kirillzyusko, line 42: I think this property can be derived from `style` property?

```tsx
style?.transform?.some(({scaleY}) => scaleY === -1);
```

src/components/KeyboardAwareScrollView/index.tsx
lpatrun: True

src/components/KeyboardAwareScrollView/index.tsx
lpatrun: Maybe I am missing how to call style, but when I call it as is in the component, I get 

"Cannot find name 'style'."

 and in rest.style I have only 
`"style": {"backgroundColor": "#FFFFFF", "flex": 1}}` so ni scaleY.

But this inverted prop should be ok because it is a Flatlist/Flashlist prop. Right?

kirillzyusko: @lpatrun can you show me how do you use `FlatList` and `KeyboardAwareScrollView`?

I assumed it would be something like that:

```tsx
<FlatList
  data={[]}
  renderItem={() => <></>}
  inverted
  renderScrollComponent={(props) => <KeyboardAwareScrollView {...props} />}
/>
```

And here (if I understand correctly) `props.style` should contain `transform` property (it should be injected by `FlatList`). Or no? 🙈

lpatrun: This is the code from the app

`
 <FlatList
                contentContainerStyle={styles.dataStyles}
                data={[...messages].reverse()}
                inverted
                renderItem={({ item }) => renderChatElement(item)}
                ItemSeparatorComponent={() => <SpacingComponent val={4} />}
                renderScrollComponent={props => <KeyboardAwareScrollView {...props} />}
                ListHeaderComponent={() => (isLoading ? <Loading /> : <></>)}
            />
 `

kirillzyusko: Thanks @lpatrun 

I'll check it today! ❤️

kirillzyusko: Hi @lpatrun 

I added this code to `KeyboardAwareScrollView` (it has TS errors):

```tsx
    const inverted = Array.isArray(rest.style)
      ? rest.style?.[0]?.transform[0]?.scaleY === -1
      : false;

    console.log("rest", rest.style, inverted);
```

And this code I used for testing:

```tsx
    <FlatList
      // contentContainerStyle={{flex: 1}}
      // style={{flexGrow: 1}}
      data={[1]}
      inverted
      renderItem={({ item }) => <></>}
      renderScrollComponent={props => <KeyboardAwareScrollView {...props} />}
  />
```

And it seems to work fine:

image

Can you check on your side again?

kirillzyusko:
@lpatrun I also merged some PRs into main and I think they can produce a conflict with your PR. May I ask you to merge/rebase your branch onto latest `main`?

kirillzyusko:
@lpatrun what is the status of this PR? Are you going to finish the changes that I requested or not? 👀

lpatrun:
@kirillzyusko my apologies, I will finish it

lpatrun:
@kirillzyusko I made changes and I have tested with and without the inverted prop.

kirillzyusko:
I approved to see if CI passes or not 👀

kirillzyusko:
@lpatrun TS is failing 😔

lpatrun:
@kirillzyusko how about now?

src/components/KeyboardAwareScrollView/index.tsx
```diff
@@ -348,21 +354,22 @@ const KeyboardAwareScrollView = forwardRef<
...
+      return enabled
+        ? {
+            // animations become choppy when scrolling to the end of the `ScrollView` (when the last input is focused)
+            // this happens because the layout recalculates on every frame. To avoid this we slightly increase padding
+            // by `+1`. In this way we assure, that `scrollTo` will never scroll to the end, because it uses interpolation
+            // from 0 to `keyboardHeight`, and here our padding is `keyboardHeight + 1`. It allows us not to re-run layout
+            // re-calculation on every animation frame and it helps to achieve smooth animation.
+            // see: https://github.com/kirillzyusko/react-native-keyboard-controller/pull/342
+            paddingBottom:
+              currentKeyboardFrameHeight.value + invertedOffset + 1,
```
kirillzyusko, line 369: Still a question about negative `extraKeyboardSpace` and `0` otherwise 🤔 

Why do we need to remove `extraKeyboardSpace` in case of inverted value?

src/components/KeyboardAwareScrollView/index.tsx
```diff
@@ -113,6 +113,12 @@ const KeyboardAwareScrollView = forwardRef<
 
     const { height } = useWindowDimensions();
 
+    const restStyle = (Array.isArray(rest.style) && rest.style?.[0]) || {};
+    const scaleStyle =
+      "transform" in restStyle &&
+      (restStyle?.transform?.[0] as { scale?: number });
+    const inverted = (scaleStyle && scaleStyle?.scale === -1) || false;
+
```
kirillzyusko, line 121: Ideally I would create a function `isInverted` (should be a pure function), I think you can put it in `utils`

And I would add `useMemo` here to avoid expensive checks 🤔 

So the code should be:

```tsx
const inverted = useMemo(() => isInverted(rest.style), [rest.style]);
```

What do you think?

kirillzyusko: @lpatrun just a friendly ping 👋

PR #669, opened 2024-10-30
Omit useNativeDriver warning on web

kubabutkiewicz:
## 📜 Description

Added check if we are on web platform to avoid warning in the console about using native driver since on web we haven't access to native driver.
## 💡 Motivation and Context

In [Expensify](https://github.com/Expensify/App) we are trying to remove all console errors and warnings in that [issue](https://github.com/Expensify/App/issues/51099). This fix will  help to achieve the goal.

### JS

- do not use `useNativeDriver` on `web`;

Tested in Expensify project.

### Before 
SCR-20241030-inlg

### After 
SCR-20241030-inzy

## 📝 Checklist

- [x] CI successfully passed

### DIFF
--- a/src/animated.tsx
+++ b/src/animated.tsx
@@ -137,7 +137,7 @@ export const KeyboardProvider = ({
             },
           },
         ],
-        { useNativeDriver: true },
+        { useNativeDriver: Platform.OS === "web" ? false : true },
       ),
     [],
   );
--- a/src/animated.tsx
+++ b/src/animated.tsx
@@ -137,7 +137,8 @@ export const KeyboardProvider = ({
             },
           },
         ],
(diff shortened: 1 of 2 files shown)


src/animated.tsx
```diff
@@ -137,7 +137,7 @@ export const KeyboardProvider = ({
             },
           },
         ],
-        { useNativeDriver: true },
+        { useNativeDriver: Platform.OS === "web" ? false : true },
```
kirillzyusko, line 140: Maybe could be simplified to `useNativeDriver: Platform.OS !== "web" `?
And add a comment above that we ignore web, because `useNativeDriver` is not supported there and just throws warnings?

kubabutkiewicz: sure let me do that

PR #681, opened 2024-11-12
feat(android): add JSPointerDispatcherCompat for RN < 0.72

gronxb:
## 📜 Description
To add a compatibility layer for `JSPointerDispatcher`, ensure that the method functions correctly regardless of whether it is called with two or three parameters.

## 💡 Motivation and Context

Related:
https://github.com/facebook/react-native/commit/1e53f88b72e24ae2654b31f5c8bfbf8a07d51e09#diff-e874545c1f508ac02d63d67356fe0519f6c0dd5f380afeb900cf1de4fce6835aL202-R203

With the change, the number of parameters has increased from two to three. Since React Native 0.71.14 only uses two parameters, this causes a build error on Android.

### Android

- add `JSPointerDispatcherCompat` class

- I've tried compiling from 0.71
- I've tried compiling from 0.76
- **I would like to test the compatibility layer to ensure it functions smoothly. Could you help me with a plan for testing this code?**

image

## 📝 Checklist

- [ ] CI successfully passed

### DIFF
--- a/android/src/main/java/com/reactnativekeyboardcontroller/views/overlay/JSPointerDispatcherCompat.kt
+++ b/android/src/main/java/com/reactnativekeyboardcontroller/views/overlay/JSPointerDispatcherCompat.kt
@@ -1,22 +1,36 @@
-
 package com.reactnativekeyboardcontroller.views.overlay
 
 import android.view.MotionEvent
 import android.view.ViewGroup
-
 import com.facebook.react.uimanager.JSPointerDispatcher
 import com.facebook.react.uimanager.events.EventDispatcher
 
 /**
  * Compat layer for `JSPointerDispatcher` interface for RN < 0.72
  */
-class JSPointerDispatcherCompat(private val viewGroup: ViewGroup) : JSPointerDispatcher(viewGroup) {
-  fun handleMotionEventCompat(event: MotionEvent?, eventDispatcher: EventDispatcher?, isCapture: Boolean) {
-    val method = try {
-      JSPointerDispatcher::class.java.getMethod("handleMotionEvent", MotionEvent::class.java, EventDispatcher::class.java, Boolean::class.javaPrimitiveType)
-    } catch (e: NoSuchMethodException) {
-      JSPointerDispatcher::class.java.getMethod("handleMotionEvent", MotionEvent::class.java, EventDispatcher::class.java)
-    }
+class JSPointerDispatcherCompat(
+  private val viewGroup: ViewGroup,
+) : JSPointerDispatcher(viewGroup) {
+  fun handleMotionEventCompat(
+    event: MotionEvent?,
+    eventDispatcher: EventDispatcher?,
+    isCapture: Boolean,
+  ) {
+    val method =
+      try {
+        JSPointerDispatcher::class.java.getMethod(
+          "handleMotionEvent",
+          MotionEvent::class.java,
+          EventDispatcher::class.java,
+          Boolean::class.javaPrimitiveType,
+        )
+      } catch (e: NoSuchMethodException) {
+        JSPointerDispatcher::class.java.getMethod(
+          "handleMotionEvent",
+          MotionEvent::class.java,
+          EventDispatcher::class.java,
+        )
+      }
     method.invoke(this, event, eventDispatcher, isCapture)
   }
 }
--- a/android/src/main/java/com/reactnativekeyboardcontroller/views/overlay/JSPointerDispatcherCompat.kt
+++ b/android/src/main/java/com/reactnativekeyboardcontroller/views/overlay/JSPointerDispatcherCompat.kt
@@ -18,6 +18,7 @@ class JSPointerDispatcherCompat(
   ) {
     val method =
       try {
+       // The RN version with 3 arguments (i.e., RN >= 0.72)
         JSPointerDispatcher::class.java.getMethod(
           "handleMotionEvent",
           MotionEvent::class.java,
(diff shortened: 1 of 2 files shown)


kirillzyusko:
thank you @gronxb for a contribution 🙌 

Let me approve and run e2e tests to see if they pass or not 👀

kirillzyusko:
> I would like to test the compatibility layer to ensure it functions smoothly. Could you help me with a plan for testing this code?

I think running e2e tests should be enough. I'm testing there that `Touchable`s from `react-native` and `react-native-gesture-handler` works as expected. So if you didn't break anything, then tests should pass - let's see 👀

gronxb:
@kirillzyusko Since ktlint was failing, I executed ktlint.

android/src/main/java/com/reactnativekeyboardcontroller/views/overlay/JSPointerDispatcherCompat.kt
```diff
@@ -0,0 +1,36 @@
...
+  ) {
+    val method =
+      try {
+        JSPointerDispatcher::class.java.getMethod(
+          "handleMotionEvent",
+          MotionEvent::class.java,
+          EventDispatcher::class.java,
+          Boolean::class.javaPrimitiveType,
+        )
+      } catch (e: NoSuchMethodException) {
```
kirillzyusko, line 27: I think `detekt` is complaining about this - we are writing `catch`, but `e` is unused (so we swallow an exception)

I think we can simply rename it to `_` (since we have only two cases at the moment) and detekt will be happy 🙂

android/src/main/java/com/reactnativekeyboardcontroller/views/overlay/JSPointerDispatcherCompat.kt
```diff
@@ -0,0 +1,36 @@
...
+        JSPointerDispatcher::class.java.getMethod(
+          "handleMotionEvent",
+          MotionEvent::class.java,
+          EventDispatcher::class.java,
+        )
+      }
+    method.invoke(this, event, eventDispatcher, isCapture)
```
kirillzyusko, line 34: Does this code works as expected (without exceptions) on RN < 0.71?

android/src/main/java/com/reactnativekeyboardcontroller/views/overlay/JSPointerDispatcherCompat.kt
gronxb: thx !

android/src/main/java/com/reactnativekeyboardcontroller/views/overlay/JSPointerDispatcherCompat.kt
```diff
@@ -0,0 +1,36 @@
...
+  private val viewGroup: ViewGroup,
+) : JSPointerDispatcher(viewGroup) {
+  fun handleMotionEventCompat(
+    event: MotionEvent?,
+    eventDispatcher: EventDispatcher?,
+    isCapture: Boolean,
+  ) {
+    val method =
+      try {
+        JSPointerDispatcher::class.java.getMethod(
```
kirillzyusko, line 21: I would add a comment which RN check we perform here (so it's always clear it's RN > 0.72 code) as first line in `try` block

android/src/main/java/com/reactnativekeyboardcontroller/views/overlay/JSPointerDispatcherCompat.kt
```diff
@@ -0,0 +1,36 @@
...
+          "handleMotionEvent",
+          MotionEvent::class.java,
+          EventDispatcher::class.java,
+          Boolean::class.javaPrimitiveType,
+        )
+      } catch (e: NoSuchMethodException) {
+        JSPointerDispatcher::class.java.getMethod(
```
kirillzyusko, line 28: And I'd also add a line here, which RN version has only 2 methods (i. e. RN <= 0.71)

android/src/main/java/com/reactnativekeyboardcontroller/views/overlay/JSPointerDispatcherCompat.kt
gronxb: https://github.com/facebook/react-native/pull/33526/files#diff-e874545c1f508ac02d63d67356fe0519f6c0dd5f380afeb900cf1de4fce6835a

If it goes to the location where the file was initially created, it should work well since the arguments will be two from the start. However, it is challenging to directly test this code. I need a scenario to test it.

gronxb: In case `getMethod` fails to retrieve the method, it likely indicates an even older version, in which case it shouldn’t execute. So, I adjusted the position of the `method.invoke` code.

kirillzyusko: Well, if you want to test your code, you can take this screen https://github.com/kirillzyusko/react-native-keyboard-controller/tree/main/example/src/screens/Examples/OverKeyboardView and do following:

- set focus to input
- wait for keyboard to appear
- press "Show" (it will show blue rectangle)
- click on blue rectangle (it should hide blue rectangle)
- click on "Show" again (it will show blue rectangle again)
- click outside of blue rectangle
- it should close it

If this scenario works on RN < 0.71, then it's safe to merge 😊

gronxb: It works well in my environment (version 0.71.14). Should I also check for compatibility with previous versions?


https://github.com/user-attachments/assets/b131dfdb-fc37-4708-9604-fc0a0999e835

kirillzyusko: @gronxb no, I don't think we need to test versions below 0.71. Thank you for testing, your PR and quick responses, I appreciate it! Let's merge this PR - code looks good for me 😊

PR #682, opened 2024-11-12
perf: lookup method only once

kirillzyusko:
## 📜 Description

In https://github.com/kirillzyusko/react-native-keyboard-controller/pull/681 we started to use reflection for detection the amount of params. But reflection is a slow mechanism, so in this PR I'm optimizing lookup stage by calculating a correct method only one time.

## 💡 Motivation and Context

Lookup can be up to 10-100 times slower comparing to the direct function call. If we call this method too frequently, we can encounter a bottleneck. To make this code faster I decided to cache the method (thus we do a lookup only one time).

This approach should almost neglect performance impact for a lookup stage. The call stage (method invocation) still will add an overhead, but it should have a small impact.

### Android

- cache method (make it lazy calculated);
- move constants to companion object.

Tested on CI.

## 📝 Checklist

- [x] CI successfully passed

### DIFF
--- a/android/src/main/java/com/reactnativekeyboardcontroller/views/overlay/JSPointerDispatcherCompat.kt
+++ b/android/src/main/java/com/reactnativekeyboardcontroller/views/overlay/JSPointerDispatcherCompat.kt
@@ -41,7 +41,7 @@ class JSPointerDispatcherCompat(
     isCapture: Boolean,
   ) {
     handleMotionEventMethod?.let { method ->
-      if (method.parameterCount == THREE_PARAMETERS) {
+      if (method.parameterCount == RN_72_PARAMS_COUNT) {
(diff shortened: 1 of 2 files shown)


gronxb:
@kirillzyusko 

https://github.com/kirillzyusko/react-native-keyboard-controller/pull/681/files#diff-e4529d4a070066ef895b72de7f05df1df3940a0c4fed960bcf0d27d9a7114651L146-R147


https://github.com/kirillzyusko/react-native-keyboard-controller/pull/681/files#diff-e4529d4a070066ef895b72de7f05df1df3940a0c4fed960bcf0d27d9a7114651L151-R154


Thank you for the refactoring! I noticed a small oversight. In that code, the second argument in the let scope should be it instead of `eventDispatcher` for `it` to be correct.

Issue #714, opened 2024-12-02
[Android 9] Crash on opening a Modal with KeyboardProvider around the app


wcastand:
**Describe the bug**
Only happening to me on <= Android 9 so far, when i open a `Modal` from `react-native`, the app crashes. error being `java.lang.RuntimeException: folly::toJson: JSON object value was an INF when serializing value at "progress"`
seems to be related to animated values hitting the INF value somehow. the crash disapear when i remove `<KeyboardController/>` from my layout.

I'm on an expo dev client app on SDK 52

Couldn't manage to a minimal reproduce so far but will keep trying, just wanted to make the issue in case the error makes more sense to you than to me x)

**Expected behavior**
No crash

**Smartphone (please complete the following information):**

- Desktop OS: MAc 15.1.1
- Device: android Samsung Galaxy S8
- OS: 9
- RN version: 0.76.3
- RN architecture: old
- JS engine: hermes
- Library version: 1.14.5

**Additional context**

here the error log:

```
ERROR  Your app just crashed. See the error below.
java.lang.RuntimeException: folly::toJson: JSON object value was an INF when serializing value at "progress"
  com.facebook.react.bridge.NativeMap.toString(Native Method)
  com.swmansion.reanimated.nativeProxy.EventHandler.receiveEvent(Native Method)
  com.swmansion.reanimated.nativeProxy.EventHandler.receiveEvent(EventHandler.java:25)
  com.facebook.react.uimanager.events.Event.dispatch(Event.java:173)
  com.swmansion.reanimated.NodesManager.handleEvent(NodesManager.java:350)
  com.swmansion.reanimated.NodesManager.onEventDispatch(NodesManager.java:336)
  com.facebook.react.uimanager.events.EventDispatcherImpl.dispatchEvent(EventDispatcherImpl.java:119)
  com.reactnativekeyboardcontroller.extensions.ThemedReactContextKt.dispatchEvent(ThemedReactContext.kt:23)
  com.reactnativekeyboardcontroller.listeners.KeyboardAnimationCallback.onProgress(KeyboardAnimationCallback.kt:267)
  androidx.core.view.WindowInsetsAnimationCompat$Impl21.dispatchOnProgress(WindowInsetsAnimationCompat.java:922)
  androidx.core.view.WindowInsetsAnimationCompat$Impl21$Impl21OnApplyWindowInsetsListener$1.onAnimationUpdate(WindowInsetsAnimationCompat.java:832)
[...]

PR #624, opened 2024-10-09
fix: incorrect `Modal` layout (if it was open while keyboard was visible)

kirillzyusko:
## 💡 Motivation and Context

Closes https://github.com/kirillzyusko/react-native-keyboard-controller/issues/619

### JS

-
-

### iOS

-
-

### Android

-
-

## 📝 Checklist

- [ ] CI successfully passed
- [ ] I added new mocks and corresponding unit-tests if library API was changed

PR #718, opened 2024-12-03
fix: crash when Modal gets shown on Android < 9

kirillzyusko:
## 📜 Description

Fixed a crash when `progress` value becomes `Infinity` and can not be serialized to JS.

## 💡 Motivation and Context

It happens when we try to divide any double/float number (except `0`) by `0.0`. In our case we were trying to divide `height / persistentKeyboardHeight`. To shed more light on the issue I modified condition `it.isNan()` to `it.isNaN() || it.isInfinite()`, and we get:

```bash
 LOG  {"duration": 160, "eventName": "onKeyboardMoveStart", "height": 0, "progress": 0, "target": -1}
 LOG  {"duration": 160, "eventName": "onKeyboardMove", "height": 47.71428680419922, "progress": 0, "target": -1}
 LOG  {"duration": 160, "eventName": "onKeyboardMove", "height": 47.71428680419922, "progress": 0, "target": -1}
 LOG  {"duration": 160, "eventName": "onKeyboardMove", "height": 47.71428680419922, "progress": 0, "target": -1}
 LOG  {"duration": 160, "eventName": "onKeyboardMove", "height": 46.57143020629883, "progress": 0, "target": -1}
 LOG  {"duration": 160, "eventName": "onKeyboardMove", "height": 42.85714340209961, "progress": 0, "target": -1}
 LOG  {"duration": 160, "eventName": "onKeyboardMove", "height": 36.57143020629883, "progress": 0, "target": -1}
 LOG  {"duration": 160, "eventName": "onKeyboardMove", "height": 28, "progress": 0, "target": -1}
 LOG  {"duration": 160, "eventName": "onKeyboardMove", "height": 17.714284896850586, "progress": 0, "target": -1}
 LOG  {"duration": 160, "eventName": "onKeyboardMove", "height": 9.142857551574707, "progress": 0, "target": -1}
 LOG  {"duration": 160, "eventName": "onKeyboardMove", "height": 2.2857143878936768, "progress": 0, "target": -1}
 LOG  {"duration": 160, "eventName": "onKeyboardMove", "height": 0, "progress": 0, "target": -1}
 LOG  {"duration": 160, "eventName": "onKeyboardMove", "height": 0, "progress": 0, "target": -1}
 LOG  {"duration": 160, "eventName": "onKeyboardMoveEnd", "height": 0, "progress": 0, "target": -1}
```

In `onStart` the `keyboardHeight=0`, `isKeyboardVisible-false` and we don't update `this.persistentKeyboardHeight` (it's 0.0 by default).

From logs above it looks like keyboard is animating from `47` pixels to `0`, but in fact there is no keyboard animation on the screen at all! 🤯 So in my understand it's a bug somewhere in google code (or in my code - but such animation is not correct). Below are the options that I considered for a fix:

### 1️⃣ add a condition, that will check if a prev `height !== 0` and if `nextHeight === 0` then ignore an animation

Such solution will add more complexity, more if-statements without an explicit benefits and warranty that it will fix/work for other similar scenarios.

### 2️⃣ check that target == -1 and ignore animation then

Not a reliable solution, because we start to depend on 3rd party variables state. And it's a known fact, that if you get a transitive dependencies, then the solution drastically becomes more complex.

### 3️⃣ add `isInfinite` check and set progress to `0` then

While it fixes a crash we still get an incorrect `height` values so any animation that depends on these values will be running and it will look strange 😬 

### 4️⃣ don't attach keyboard callbacks to Modal window on Android 9

The most perspective solution - on Android < 12 all events are passing through the main window (and actually attaching a callback is not necessary there, but when I implemented https://github.com/kirillzyusko/react-native-keyboard-controller/pull/466 and https://github.com/kirillzyusko/react-native-keyboard-controller/pull/624 I decided to keep the code for consistency). So if we remove these callbacks (if we attach them to window only for Android 12) then all code works well on both platforms.

Closes https://github.com/kirillzyusko/react-native-keyboard-controller/issues/714

### Android

- don't attach keyboard callbacks on Android < 12 to `Modal` window;

Tested in reproduction example and on CI.

https://github.com/user-attachments/assets/76331e0d-6779-4a4b-a837-ba4802ac4139

## 📝 Checklist

- [x] CI successfully passed
- [x] I added new mocks and corresponding unit-tests if library API was change

### DIFF
--- a/android/src/main/java/com/reactnativekeyboardcontroller/modal/ModalAttachedWatcher.kt
+++ b/android/src/main/java/com/reactnativekeyboardcontroller/modal/ModalAttachedWatcher.kt
@@ -71,7 +71,8 @@ class ModalAttachedWatcher(
         // go through main `rootView`, so we don't need to stop main
         // callback - otherwise keyboard transitions will not be animated
         this.callback()?.suspend(true)
-        // attaching callback to Modal on Android < 12 can cause ghost animation, see: https://github.com/kirillzyusko/react-native-keyboard-controller/issues/714
+        // attaching callback to Modal on Android < 12 can cause ghost animations, see: https://github.com/kirillzyusko/react-native-keyboard-controller/pull/718
+        // and overall attaching additional callbacks (if animation events go through the main window) is not necessary
         ViewCompat.setWindowInsetsAnimationCallback(rootView, callback)
         ViewCompat.setOnApplyWindowInsetsListener(eventView, callback)
 
--- a/android/src/main/java/com/reactnativekeyboardcontroller/modal/ModalAttachedWatcher.kt
+++ b/android/src/main/java/com/reactnativekeyboardcontroller/modal/ModalAttachedWatcher.kt
@@ -65,14 +65,16 @@ class ModalAttachedWatcher(
         )
 
       rootView.addView(eventView)
-      // on Android < 12 all events for `WindowInsetsAnimationCallback`
-      // go through main `rootView`, so we don't need to stop main
-      // callback - otherwise keyboard transitions will not be animated
-      this.callback()?.suspend(areEventsComingFromOwnWindow)
-      ViewCompat.setWindowInsetsAnimationCallback(rootView, callback)
-      ViewCompat.setOnApplyWindowInsetsListener(eventView, callback)
 
       if (areEventsComingFromOwnWindow) {
+        // on Android < 12 all events for `WindowInsetsAnimationCallback`
+        // go through main `rootView`, so we don't need to stop main
+        // callback - otherwise keyboard transitions will not be animated
+        this.callback()?.suspend(true)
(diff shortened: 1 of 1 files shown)


antFrancon:
I'm not sure if the root cause is the same, but I encountered some crashes in production from the same piece of code (`height / persistentKeyboardHeight` resulting in INF). This issue isn't limited to Android versions below 9.
image

kirillzyusko:
@antFrancon may I ask you to update to `1.15.0` and check if the problem happens again? If yes, then please, open a new issue with all the information that you can gather and we can decide what to do next to fix this problem 👀 

Maybe you can discover when this problem happens, i. e. user went from background to foreground, or opened a Modal or something else - any additional information can be super helpful 🙏

antFrancon:
@kirillzyusko Of course! I'm investigating this right now. I'll open a separate issue after gathering more information.

kirillzyusko:
@antFrancon thank you! ❤️

Issue #739, opened 2024-12-19
Keyboard animation progress may be infinite and crash app


matthieuMay:
**Describe the bug**
In production we randomly see a crash with the following stack trace
java.lang.RuntimeException: folly::toJson: JSON object value was an INF when serializing value at "progress"
    at com.facebook.react.bridge.NativeMap.toString(NativeMap.kt)
    at com.swmansion.reanimated.nativeProxy.EventHandler.receiveEvent(EventHandler.java)
    at com.swmansion.reanimated.nativeProxy.EventHandler.receiveEvent(EventHandler.java:25)
    at com.facebook.react.uimanager.events.Event.dispatch(Event.java:173)
    at com.swmansion.reanimated.NodesManager.handleEvent(NodesManager.java:350)
    at com.swmansion.reanimated.NodesManager.onEventDispatch(NodesManager.java:336)
    at com.facebook.react.uimanager.events.EventDispatcherImpl.dispatchEvent(EventDispatcherImpl.java:119)
    at com.reactnativekeyboardcontroller.extensions.ThemedReactContextKt.dispatchEvent(ThemedReactContext.kt:23)
    **at com.reactnativekeyboardcontroller.listeners.KeyboardAnimationCallback.onProgress(KeyboardAnimationCallback.kt:267)**
    at androidx.core.view.WindowInsetsAnimationCompat$Impl21.dispatchOnProgress(WindowInsetsAnimationCompat.java:922)
    at androidx.core.view.WindowInsetsAnimationCompat$Impl21$Impl21OnApplyWindowInsetsListener$1.onAnimationUpdate(WindowInsetsAnimationCompat.java:832)
    at android.animation.ValueAnimator.animateValue(ValueAnimator.java:1558)
    at android.animation.ValueAnimator.animateBasedOnTime(ValueAnimator.java:1349)
    at android.animation.ValueAnimator.doAnimationFrame(ValueAnimator.java:1481)
    at android.animation.AnimationHandler.doAnimationFrame(AnimationHandler.java:146)
    at android.animation.AnimationHandler.access$100(AnimationHandler.java:37)
    at android.animation.AnimationHandler$1.doFrame(AnimationHandler.java:54)
    at android.view.Choreographer$CallbackRecord.run(Choreographer.java:1278)
    at android.view.Choreographer.doCallbacks(Choreographer.java:1019)
    at android.view.Choreographer.doFrame(Choreographer.java:907)
[...]

Issue #737, opened 2024-12-18
Crash when focusing on an input


bmorales-broditec:
**Describe the bug**
I'm using KeyboardProvider and KeyboardToolbar to wrap screens that use inputs, when focusing on an input my application crashes and closes. This error happens often, but not always.

![Screenshot 2024-12-18 at 10 19 48 AM](https://github.com/user-attachments/assets/6f65fb6e-b833-45a8-abe6-09a407871dd0)

https://github.com/user-attachments/assets/de5178b1-2264-40dc-9470-d70b8fd91652

**Code snippet**
```typescript
import React from 'react';

import {KeyboardProvider, KeyboardToolbar} from 'react-native-keyboard-controller';

import {KeyboardHandlerProps} from './keyboard-handler.interface';
import {useTranslation} from '@/shared/locales';

export const KeyboardHandler = ({children, showNavigation = false}: KeyboardHandlerProps) => {
  const {getLocalizedText} = useTranslation();

  return (
    <KeyboardProvider>
      {children}
      <KeyboardToolbar doneText={getLocalizedText('done')} showArrows={showNavigation} />
    </KeyboardProvider>
  );
};
```

Example:

```typescript
    <KeyboardHandler showNavigation>
      <SafeAreaContainer contentContainerStyle={styles.container} edges={['top']}>
        <View style={styles.content}>
          <Typography color="purpleMain" style={styles.title} variant="xxlBold">
            {getLocalizedText('changeYourPassword')}
          </Typography>
          <Typography color="grayDark" style={styles.description} variant="mRegular">
            {getLocalizedText('secureYourAccountWithANewPassword')}
          </Typography>

          <Controller
            control={control}
            name="password"
            render={({field: {onChange, onBlur, value}}) => (
              <View style={styles.changeTemporalPasswordFieldContainer}>
                <PasswordField
                  containerStyle={styles.changeTemporalPasswordField}
                  error={!!errors?.password}
                  helperText={errors?.password?.message}
                  label={getLocalizedText('password')}
                  maxLength={50}
                  placeholder={getLocalizedText('password')}
                  required
                  showStrongPasswordIndicators
                  value={value}
                  onBlur={onBlur}
                  onChangeText={onChange}
                />
              </View>
            )}
            rules={{
              validate: {
                required: v => requiredValidation(v) || getLocalizedText('requiredField'),
                isValidPassword: v => isValidPassword(v),
              },
            }}
          />
[...]

Issue #735, opened 2024-12-17
Crash when modal gets shown 


exzos28:
The issue is still valid.
The project has:
- react-native-keyboard-controller
- @gorhom/bottom-sheet

When I open a modal window I get a crash
`java.lang.RuntimeException: folly::toJson: JSON object value was an INF when serializing value at “progress”`

A bit of information from the debugger:
<img width="1186" alt="image" src="https://github.com/user-attachments/assets/18f5a55e-1421-4906-a889-3eddecdd1ba1" />

- OS: android 10
- RN version: 0.76
- RN architecture: new, fabric
- JS engine: Hermes
- Library version: 1.15.0

**Additional context**
#718
#714

exzos28:
progress is INF because of persistentKeyboardHeight, it's 0.0 
Do you really need a reproducible repository? Let me know if there's anything I can do to help here.

exzos28:
Can we just check if it isn't infinity?

`progress = abs((height / persistentKeyboardHeight)).let { if (it.isNaN() || it.isInfinite()) 0.0 else it }`

kirillzyusko:
> Do you really need a reproducible repository? 

Well, if you can prepare it that would be a dope!

> Can we just check if it is infinity?

We can and it will fix a crash, but I would like to use it only as a last resort. If we add that check we'll send `progress=0` and `height=45` (for example), which will add inconsistency - if you do `height`- based animations you will see that animation, 
 and if you use `progress`, then no 😔 🤷‍♂️ 

So my assumption is that we need to figure out why those insets changes are coming 🤔 Do you think it's doable to create a reproduction example?

kirillzyusko:
Okay @exzos28 

I also see that a similar issue has been created in https://github.com/kirillzyusko/react-native-keyboard-controller/issues/737

I feel a pain, when the app crashes, so I think it makes sense to add a dirty workaround with converting a value to `0` to avoid crash. It's better to have an incorrect animation for end user rather than a crash of the application 😔

kirillzyusko:
I merged a fix - at least it should prevent a crash (but may lead to unexpected/inconsistent UI state: https://github.com/kirillzyusko/react-native-keyboard-controller/pull/740#issuecomment-2553315695

Let me know if you encounter any further issues. If you can prepare a reproduction example that would be a dope, since we can manage to discover a root problem and provide a fix 🤞

PR #740, opened 2024-12-19
fix: handle progress being infinite

matthieuMay:
## 📜 Description

Add a numerical safety test to keyboard animation progress

## 💡 Motivation and Context

It solves a random crash in production where progress is infinite

Fixes https://github.com/kirillzyusko/react-native-keyboard-controller/issues/739 https://github.com/kirillzyusko/react-native-keyboard-controller/issues/737 https://github.com/kirillzyusko/react-native-keyboard-controller/issues/735

### Android

- added `isInfinite` check along with `isNaN`;

Releasing it to production via a patch package the error report disappeared

## 📝 Checklist

- [x] CI successfully passed

### DIFF
--- a/android/src/main/java/com/reactnativekeyboardcontroller/listeners/KeyboardAnimationCallback.kt
+++ b/android/src/main/java/com/reactnativekeyboardcontroller/listeners/KeyboardAnimationCallback.kt
@@ -251,7 +251,7 @@ class KeyboardAnimationCallback(
 
     var progress = 0.0
     try {
-      progress = abs((height / persistentKeyboardHeight)).let { if (it.isNaN()) 0.0 else it }
+      progress = abs((height / persistentKeyboardHeight)).let { if (it.isNaN() || it.isInfinite()) 0.0 else it }
     } catch (e: ArithmeticException) {
       // do nothing, just log an exception send progress as 0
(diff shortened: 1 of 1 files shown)


kirillzyusko:
A small note from the author of this lib - such check will lead to UI inconsistency (because we are trying to divide non-zero `height` by `0`, so with this fix we'll send something like `{ height: 40, progress: 0 }`). If you depend on `height` value, you may see a "ghost" keyboard animations (when there is actually no keyboard on a screen, but you see the part of the animation).

I understand, that it may be hard to prepare a reproduction, because it seems like the error is very flaky and it can be present only in heavy prod apps with reach functionality (i. e. it can be a race conditions and it's hard to reproduce it in simple reproduction example).

To deliver a better user experience and don't crash app in case if value can not be serialized - you convinced me guys to have that check in the codebase. It's better to show fake/ghost animation rather than crash the app and totally block user interaction with the app.

Still hope that at some point of time someone will prepare a reproduction example and we can figure out what exactly causes this problem. For now - let's have this fix in a codebase to avoid app crashes 🙂

kirillzyusko:
And yes, seems like this problem affects many users, so i'll merge a fix and will publish `1.15.1` version straight away 😊

PR #755, opened 2025-01-14
fix: selection coordinates on iOS

kirillzyusko:
## 💡 Motivation and Context

image

image

### JS

-
-

### iOS

-
-

### Android

-
-

## 📝 Checklist

- [ ] CI successfully passed


kirillzyusko:
According to e2e tests we over-specify the `y`-coordinate, so most likely the chosen approach is incorrect, so I'll close the PR. We can return to it at any time later on 👀

Issue #592, opened 2024-09-22
Inconsistent `NavigationBar` Behavior with `Edge-to-Edge` Support (Android)


wise-danya:
### Description
On Android, when toggling the `enabled` prop of the `<KeyboardProvider />` using `useKeyboardController`, the system `NavigationBar` behaves inconsistently due to the handling of `edge-to-edge` mode:

When `enabled` is set to `true`, the `NavigationBar` disappears.
When `enabled` is set to `false`, the `NavigationBar` reappears and causes layout issues by lifting the content.
This issue is primarily caused by toggling `edge-to-edge` mode in the `enable()` and `disable()` methods.

### Steps to reproduce the behavior

1. Integrate `react-native-keyboard-controller` and `react-native-bars` into a project.
2. Use the `useKeyboardController` hook to toggle the enabled prop of `<KeyboardProvider />`.
3. Toggle `enabled` to `true`, observe that the system `NavigationBar` disappears.
4. Toggle `enabled` to `false`, observe that the `NavigationBar` reappears and lifts the content.

### Expected behavior
The `NavigationBar` should remain hidden or translucent and should not affect the layout or content positioning, regardless of the enabled state of the `KeyboardProvider`.

### Repo for reproducing

You can find a minimal reproducible example for this issue in the following repository: https://github.com/wise-danya/rn-keyboard-controller-issue-example

### Video Demo

https://github.com/user-attachments/assets/b24e775e-021a-4131-ba13-bd29a497cf73

### Additional context
On some screens, when the library is disabled, the `NavigationBar` is visible and causes the content to be lifted, leading to layout issues. This makes it hard to gradually adopt the library across multiple screens in an app.

### Environment:

- Desktop OS: MacOS 14.6.1
- Device: Pixel 7
- OS: Android 14
- RN version: Example: 0.75.3; Prod: 0.73.9
- RN architecture: New Architecture (Fabric)
- JS engine: Hermes
- Library version: 1.13.4

kirillzyusko:
Hey @wise-danya 👋 Thank you for providing a repo and high quality issue description!

Before digging into internals may I ask you what is the purpose of disabling the module and keeping edge-to-edge mode?

[...]

PR #765, opened 2025-01-19
Add preserveEdgeToEdge prop

zoontek:
## 📜 Description

Currently, even when `react-native-edge-to-edge` is installed, RNKC is able to disable `edge-to-edge`. For example, if we set `enable={false}` on `KeyboardProvider`, `disable` will be called on `EdgeToEdgeReactViewGroup`, and `WindowCompat.setDecorFitsSystemWindows` too.

This change detect if `react-native-edge-to-edge` is installed. If it is, edge-to-edge is forced and cannot be disabled.

## 💡 Motivation and Context

It fixes an issue that breaks edge-to-edge (not only when enabled with `react-native-edge-to-edge`, but in general).

Fixes https://github.com/kirillzyusko/react-native-keyboard-controller/issues/592 (_Toggle enabled to false, observe that the NavigationBar reappears and lifts the content._)

### Android

- Fixes issue with edge-to-edge when `KeyboardProvider` is used with `enabled={false}`.

- Using the example app (`react-native-edge-to-edge` installed). Before, when switching on the "14. Enable / disable" screen, we were able to disable edge-to-edge too, not only keyboard management. Now it stays edge-to-edge.

Screenshot 2025-01-19 at 22 59 31

## 📝 Checklist

- [x] CI successfully passed

### DIFF
--- a/android/src/fabric/java/com/reactnativekeyboardcontroller/KeyboardControllerViewManager.kt
+++ b/android/src/fabric/java/com/reactnativekeyboardcontroller/KeyboardControllerViewManager.kt
@@ -36,6 +36,12 @@ class KeyboardControllerViewManager(
     value: Boolean,
   ) = manager.setNavigationBarTranslucent(view as EdgeToEdgeReactViewGroup, value)
 
+  @ReactProp(name = "preserveEdgeToEdge")
+  override fun setPreserveEdgeToEdge(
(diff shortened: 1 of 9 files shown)


zoontek:
@kirillzyusko Done!

kirillzyusko:
Thank you for your contribution @zoontek ❤️

PR #770, opened 2025-01-21
feat: add ability to turn off adding extra padding for keyboard

xotahal:
## 📜 Description

Added an option to specify whether to include keyboard height padding. This is particularly useful for scenarios like bottom sheets, where the position of the modal is already managed relative to the keyboard, and additional padding is unnecessary.

## 💡 Motivation and Context

This change resolves the issue of redundant keyboard padding in components like bottom sheets, which already handle modal positioning relative to the keyboard. By providing an option to disable the keyboard height padding, developers have more control over the layout and can avoid unwanted spacing.

### JS

- Added a new property to control whether to add keyboard height padding.

### iOS

- No platform-specific changes.

### Android

- No platform-specific changes.

Tested by integrating the new property with a bottom sheet component to ensure that keyboard height padding is correctly applied or excluded based on the property value. Checked the behavior on both iOS and Android devices.

### Before:

https://github.com/user-attachments/assets/fb4a5a38-d410-42b7-b34c-e174ac421c23

### After:

https://github.com/user-attachments/assets/5e3c89d4-f001-4b13-a4bb-276e51c8c2c1

## 📝 Checklist

- [ ] CI successfully passed

### DIFF
--- a/docs/docs/api/components/keyboard-aware-scroll-view.mdx
+++ b/docs/docs/api/components/keyboard-aware-scroll-view.mdx
@@ -115,6 +115,10 @@ If you have _**sticky elements**_ above the keyboard and want to extend the keyb
 This property acts as [extraScrollHeight](https://github.com/APSL/react-native-keyboard-aware-scroll-view/tree/9eee405f7b3e261faf86a0fc8e495288d91c853e?tab=readme-ov-file#props) from original [react-native-keyboard-aware-scroll-view](https://github.com/APSL/react-native-keyboard-aware-scroll-view) package.
 :::
 
+### `withKeyboardHeightPadding`
+
+Specifies whether to add padding equal to the keyboard height. This is useful when working with components like a bottom sheet, where the bottom sheet manages the position of the modal above the keyboard. In such cases, adding extra padding for the keyboard is unnecessary.
+
 ## Integration with 3rd party components
 
 ### `FlatList`/`FlashList`/`SectionList` etc.
--- a/src/components/KeyboardAwareScrollView/index.tsx
+++ b/src/components/KeyboardAwareScrollView/index.tsx
@@ -39,6 +39,8 @@ export type KeyboardAwareScrollViewProps = {
   enabled?: boolean;
   /** Adjusting the bottom spacing of KeyboardAwareScrollView. Default is `0` */
   extraKeyboardSpace?: number;
+  /** Specifies whether to add padding equal to the keyboard height. This is useful when working with components like a bottom sheet, where the bottom sheet manages the position of the modal above the keyboard. In such cases, adding extra padding for the keyboard is unnecessary. */
+  withKeyboardHeightPadding?: boolean;
   /** Custom component for `ScrollView`. Default is `ScrollView` */
   ScrollViewComponent?: React.ComponentType<ScrollViewProps>;
 } & ScrollViewProps;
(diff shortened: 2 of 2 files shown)


xotahal:
Hi @kirillzyusko,

Thanks for the quick reply! I really appreciate your time and help 🙏






I’ve uploaded an example that will hopefully clarify the issue a bit better. I’m working with a bottom sheet that contains a “long” form and isn’t detached from the bottom edge. The bottom sheet essentially covers the whole screen, making it feel more like a regular screen with a long form.

Before:

https://github.com/user-attachments/assets/2328a8a4-5db8-492a-8144-20f8251c6425

After:

https://github.com/user-attachments/assets/c95c21dc-4780-49d7-a90b-93104df301db

To make scrolling work smoothly within the bottom sheet, I’m relying on the scrollables provided by the library (https://gorhom.dev/react-native-bottom-sheet/scrollables). This approach works fine and is understandable. Following your suggestion, I’m also trying to use the KeyboardAwareScrollView in combination with the bottom sheet library (https://kirillzyusko.github.io/react-native-keyboard-controller/docs/api/components/keyboard-aware-scroll-view#gorhombottom-sheet).

First, let me say how much I like your library! It’s incredibly reliable and offers features that other libraries don’t, which I find essential for my use case.

That said, the challenge arises because the bottom sheet library also has its own way of handling keyboard avoidance. If I only rely on their approach, I run into some of the issues you’ve outlined in your docs (https://kirillzyusko.github.io/react-native-keyboard-controller/docs/api/components/keyboard-aware-scroll-view#comparison), and I lose access to some of the valuable features your library provides.

On the other hand, if I only use your KeyboardAwareScrollView, I run into a different problem. When the bottom sheet is smaller than the keyboard (for example, in compact states), your library adds padding to the scroll view, but the entire bottom sheet itself doesn’t move above the keyboard—it stays behind it. So in this scenario, I still need the bottom sheet library’s keyboard-handling logic to push the sheet above the keyboard.

See this example:

https://github.com/user-attachments/assets/2329de91-3351-4ccd-b447-9a4f8285ad8e




Essentially, I need to combine the strengths of both libraries to get the desired behavior. However, this creates an issue with “double” padding, where both libraries add padding at the same time, which isn’t what I need.

Hopefully, this explanation makes at least a little sense. I truly appreciate all the hard work you’ve put into this library—it’s been a game-changer in so many ways! If you have any suggestions or ideas for tackling this situation, I’d be grateful for your guidance.

Thanks again!

hirbod:
The thread is super long, so I didn't read it. What we done was to disable all the Keyboard handling that was shipped by Gorhom and manage everything with this library.

```tsx
  // this moves the bottom sheet up when the keyboard is open
    const viewStyle = useAnimatedStyle(() => {
      return {
        transform: [{ translateY: shouldAdjustForKeyboard && isFocused ? height.value : 0 }],
      }
    })

    return (
        <AnimatedView className="flex-grow" style={viewStyle}>
          <GestureHandlerRootView className="flex-grow">
            <BottomSheet ....
```

Kjagd:
I am currently migrating to this (very helpful!) library, and I made the mistake of not changing the `BottomSheetTextInput` to a regular `TextInput` which indeed caused both libraries to manage the sizing at the same time.

https://gorhom.dev/react-native-bottom-sheet/components/bottomsheettextinput

PR #779, opened 2025-01-25
perf: update perf metrics

kirillzyusko:
### JS

-
-

### iOS

-
-

### Android

-
-

## 📝 Checklist

- [ ] CI successfully passed


kirillzyusko:
Close as it seems like runners migrated to a different hardware and now baseline always get re-created on CI.

Issue #847, opened 2025-03-08
Allow us to hide the done button on the KeyboardToolbar


spaansba:
First of all a big thank you for maintaining this project, the new KeyboardToolbar is exactly what I wanted!!

**Describe the solution you'd like**
A way to hide the "done" button in the KeyboardToolbar just like we can hide the arrows with the showArrows prop. The perfect solution would be this:

```
 <KeyboardToolbar
        offset={{ opened: tabBarHeight }}
        showArrows={false}
        showDoneButton={false} // defaults to true like showArrows
 />
```

**Describe alternatives you've considered**
I tried setting the dontText to an empty string but it rightfully returns "done"

PR #849, opened 2025-03-11
feat: ability to hide "Done" button in `KeyboardToolbar`

kirillzyusko:
## 📜 Description

Don't show "Done" button, if text is empty.

## 💡 Motivation and Context

If people want to hide "Done" button it may be intuitive guess that `doneText` should be empty text.

I don't want to introduce more props, because `KeyboardToolbar` already has a plenty of them, so let's make `doneText` behavior more predictable.

Closes https://github.com/kirillzyusko/react-native-keyboard-controller/issues/847

### JS

- assign "Done" as default `doneText` instead of using `||` operator.

Tested on iPhone 15 Pro (iOS 17.5), fabric.

image

## 📝 Checklist

- [x] CI successfully passed


spaansba:
Thanks for implementing this!

Issue #848, opened 2025-03-08
KeyboardToolbar still shows container of arrows even when showArrows={false}


spaansba:
**Describe the bug**
When you set showArrows={false} in the KeyboardToolbar component, you would expect the arrows to be completely removed from the UI. However, while the arrows themselves disappear, their container wrapper remains visible, leaving an empty space with the background color still showing on the left side of the custom toolbar.

**Code snippet**

```
<KeyboardToolbar
      offset={{ opened: tabBarHeight }}
      showArrows={false}
      content={<View className="size-full bg-red-300"></View>}
    />
```

**Repo for reproducing**
https://github.com/spaansba/Toaster
using it in components/messenger/MessengerWrapper

**To Reproduce**
1. Implement the KeyboardToolbar with showArrows={false}
2. Run the app on a device or simulator
3. Open the keyboard to see the toolbar
4. Notice the empty space on the left side where arrows would normally be

**Expected behavior**
When setting showArrows={false}, the entire arrow section (including its container) should be removed from the layout, with no empty space remaining.

**Smartphone (please complete the following information):**

- Device: iPhone 13
- OS: IOS 18.3.1
- RN version: 0.76.7
- expo version: 52.0.35
- RN architecture: new
- Library version: 1.16.0

**Screenshots**
![Image](https://github.com/user-attachments/assets/250540e2-3c25-4fa0-8286-fc9349abc5d9)

kirillzyusko:
@spaansba i.e. you want your pink view to touch left border of the screen, right?

spaansba:
Exactly @kirillzyusko

PR #851, opened 2025-03-11
Removed left padding on KeyboardToolbar when navigation arrows are hidden

spaansba:
## 📜 Description
Remove padding when showArrows={false} on KeyboardToolbar

The KeyboardToolbar component currently applies a fixed paddingHorizontal of 8px even when showArrows is set to false. This creates an unnecessary left padding gap when arrows are hidden. This PR modifies the toolbar styling to conditionally apply paddingLeft only when arrows are visible.

## 💡 Motivation and Context

When setting showArrows={false}, users expect the whole container to be gone. 

Closes https://github.com/kirillzyusko/react-native-keyboard-controller/issues/848

### JS
- Removed paddingHorizontal from toolbar styles
- Added conditional paddingLeft based on showArrows prop
- Added paddingRight to maintain right-side spacing
- Added showArrows to toolbarStyle useMemo dependencies

Tested manually with both showArrows={true} and showArrows={false} to ensure:
- With arrows visible, padding is properly applied on both sides (8px)
- With arrows hidden, left padding is removed (0px) while right padding is maintained (8px)

Before: 
Before

After:
After

## 📝 Checklist

- [X] CI successfully passed

### DIFF
--- a/src/components/KeyboardToolbar/index.tsx
+++ b/src/components/KeyboardToolbar/index.tsx
@@ -109,9 +109,11 @@ const KeyboardToolbar: React.FC<KeyboardToolbarProps> = ({
       styles.toolbar,
       {
         backgroundColor: `${theme[colorScheme].background}${opacity}`,
+        paddingLeft: showArrows ? 8 : 0,
+        paddingRight: 8,
       },
     ],
-    [colorScheme, opacity, theme],
+    [colorScheme, opacity, theme, showArrows],
   );
   const offset = useMemo(
     () => ({ closed: closed + KEYBOARD_TOOLBAR_HEIGHT, opened }),
@@ -222,7 +224,6 @@ const styles = StyleSheet.create({
     width: "100%",
     flexDirection: "row",
     height: KEYBOARD_TOOLBAR_HEIGHT,
-    paddingHorizontal: 8,
   },
   doneButton: {
     fontWeight: "600",
--- a/src/components/KeyboardToolbar/index.tsx
+++ b/src/components/KeyboardToolbar/index.tsx
@@ -225,7 +225,6 @@ const styles = StyleSheet.create({
   },
   arrowsButton: {
     flexDirection: "row",
-    gap: 5,
     paddingHorizontal: DEFAULT_BUTTON_PADDING,
   },
   doneButton: {
--- a/src/components/KeyboardToolbar/index.tsx
+++ b/src/components/KeyboardToolbar/index.tsx
@@ -64,6 +64,7 @@ const TEST_ID_KEYBOARD_TOOLBAR_DONE = `${TEST_ID_KEYBOARD_TOOLBAR}.done`;
 
 const KEYBOARD_TOOLBAR_HEIGHT = 42;
 const DEFAULT_OPACITY: HEX = "FF";
+const DEFAULT_BUTTON_PADDING = 8;
 
 /**
  * `KeyboardToolbar` is a component that is shown above the keyboard with `Prev`/`Next` and
@@ -109,11 +110,9 @@ const KeyboardToolbar: React.FC<KeyboardToolbarProps> = ({
       styles.toolbar,
       {
         backgroundColor: `${theme[colorScheme].background}${opacity}`,
-        paddingLeft: showArrows ? 8 : 0,
-        paddingRight: 8,
       },
     ],
-    [colorScheme, opacity, theme, showArrows],
+    [colorScheme, opacity, theme],
   );
   const offset = useMemo(
     () => ({ closed: closed + KEYBOARD_TOOLBAR_HEIGHT, opened }),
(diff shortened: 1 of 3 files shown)


src/components/KeyboardToolbar/index.tsx
```diff
@@ -109,9 +109,11 @@ const KeyboardToolbar: React.FC<KeyboardToolbarProps> = ({
       styles.toolbar,
       {
         backgroundColor: `${theme[colorScheme].background}${opacity}`,
+        paddingLeft: showArrows ? 8 : 0,
+        paddingRight: 8,
```
kirillzyusko, line 113: My current idea was to add `View` instead of `<>` for button group and apply a style `paddingLeft` there.

And accordingly I wanted to remove `paddingHorizontal` from container and add `paddingRight: 8` to `doneButtonContainer`. In this case we will not need to write additional conditional code 🤔

kirillzyusko: Can you try to make these changes and I'll run CI to see if it passes or not? 👀

spaansba: That was exactly my first idea as well! 

What I think might be an even better approach is the following:

- Remove `paddingHorizontal: 8` from `toolbar `style
- Add a `View ` instead of `<>` and give it the following styles: 
```
  arrowsButton: {
    flexDirection: 'row',
    gap: 5,
    paddingHorizontal:8,
  },
```

- Remove the done `doneButtonContainer ` styles
- add `paddingHorizonal: 8` to `doneButton`
- in Arrow.tsx remove `marginHorizontal: 5` from `arrowUpContainer`

This way both done and arrow buttons have a padding of 8 and the 2 arrow buttons still have the gap of 5 between them. Also with this if the arrow or done button gets removed the whole container is removed as well.

Also in the previous version the done button didnt have padding on the left only on the right (margin). Now it is centered in its container.

src/components/KeyboardToolbar/index.tsx
```diff
@@ -222,14 +222,18 @@ const styles = StyleSheet.create({
     width: "100%",
     flexDirection: "row",
     height: KEYBOARD_TOOLBAR_HEIGHT,
-    paddingHorizontal: 8,
+  },
+  arrowsButton: {
+    flexDirection: "row",
+    gap: 5,
```
kirillzyusko, line 228: I think `gap` was added relatively recently and I support older RN versions as well, so it's better to use paddings/margins to make sure all library users have consistent experience 😊

src/components/KeyboardToolbar/index.tsx
```diff
@@ -222,14 +222,18 @@ const styles = StyleSheet.create({
...
+    paddingHorizontal: DEFAULT_BUTTON_PADDING,
   },
   doneButton: {
+    alignItems: "center",
+    justifyContent: "center",
```
kirillzyusko, line 233: Why it's needed now? 🤔

src/components/KeyboardToolbar/index.tsx
```diff
@@ -222,14 +222,18 @@ const styles = StyleSheet.create({
...
     fontWeight: "600",
     fontSize: 15,
-  },
-  doneButtonContainer: {
-    marginRight: 8,
+    paddingHorizontal: DEFAULT_BUTTON_PADDING,
```
kirillzyusko, line 236: With a new approach arrow both done/arrows will have a horizontal padding, right?

I think with a previous approach only arrows had `paddingRight` and theoretically long content could overlap with "Done" button (since Done buttong didn't have any margin/padding from left). I think a new approach makes sense 👍

src/components/KeyboardToolbar/index.tsx
spaansba: Ah you are right, thats not needed.

src/components/KeyboardToolbar/index.tsx
spaansba: yes with this approach both containers will have the exact amount of horizontal padding. 

I tested it with long content (a custom toolbar) and there is no overlapping in this approach with or without the arrows/done button!

src/components/KeyboardToolbar/index.tsx
spaansba: Makes sense! I remove the `gap ` and added `marginRight: 2.5` and `marginLeft: 2.5` on the arrowUp/down containers respectively. I also made it so the arrowContainer is a base container and the arrowUp/down add styles to the base container instead of the down container adding styles to the up container.

kirillzyusko:
@spaansba after your changes button have different positioning:

ToolbarFirstInputFocused-diff

This component is a pretty precise replica of iOS toolbar and I'd like to continue to have pixel-perfect UI 🙂

spaansba:
interesting, Ill have a look

kirillzyusko:
Closing in favour of https://github.com/kirillzyusko/react-native-keyboard-controller/pull/865

Issue #470, opened 2024-06-13
KeyboardToolbar's button click will suddenly open the BottomSheet and immediately focus on the BottomSheetTextInput inside the BottomSheet


7dp:
Hi 🙌. Thank you so much for your hard work on this amazing lib. 
Truly useful for the community. ❤️

**Describe the bug**
So i have a screen in my RN app. The screen has a several `TextInput` inside a `KeyboardAwareScrollView` (KAS is from this lib). And i have a `BottomSheet` (from Gorhom) that has a `BottomSheetTextInput` inside it. Lastly, i have a `KeyboardToolbar`. 
The problem begin when i focused to the last `TextInput` inside KAS and then i click the next button of `KeyboardToolbar` and then suddenly my `BottomSheet` (that i haven't opened yet before) is appear with the `BottomSheetTextInput` is focused.

**Code snippet**
```typescript
function App(): React.JSX.Element {
  const sheetRef = useRef<RNBottomSheet>(null);

  return (
    <GestureHandlerRootView style={styles.flex}>
      <KeyboardProvider>
        <SafeAreaView style={[styles.background, styles.flex]}>
          <KeyboardAwareScrollView
            contentContainerStyle={styles.kasContentContainer}
            keyboardShouldPersistTaps="handled"
            bottomOffset={keyboardBottomOffset}>
            {Array(10)
              .fill('')
              .map((_item, index) => (
                <TextInput
                  key={index}
                  placeholderTextColor="lightgray"
                  placeholder="Type here..."
                  style={styles.input}
                />
              ))}
            <Button
              title="Open Sheet"
              onPress={() => {
                sheetRef.current?.expand();
              }}
            />
          </KeyboardAwareScrollView>
          <KeyboardToolbar />
          <BottomSheet ref={sheetRef}>
            <View style={styles.container}>
              <BottomSheetTextInput
                placeholderTextColor="lightgray"
                placeholder="Type here..."
                style={styles.input}
              />
            </View>
          </BottomSheet>
        </SafeAreaView>
      </KeyboardProvider>
    </GestureHandlerRootView>
  );
}
```

**Repo for reproducing**
https://github.com/7dp/KeyboardToolbarBottomSheetRepro

**To Reproduce**

1. Focus the keyboard on the last `TextInput` inside KAS.
2. Click next button of `KeyboardToolbar`.
3. `BottomSheet` will suddenly appear with a focused `BottomSheetTextInput`.

**Expected behavior**
[...]

PR #881, opened 2025-03-25
feat: `KeyboardToolbar.Exclude`

kirillzyusko:
## 💡 Motivation and Context

Closes https://github.com/kirillzyusko/react-native-keyboard-controller/issues/470

### JS

-
-

### iOS

-
-

### Android

-
-

## 📝 Checklist

- [ ] CI successfully passed

### DIFF
--- a/android/src/main/java/com/reactnativekeyboardcontroller/traversal/ViewHierarchyNavigator.kt
+++ b/android/src/main/java/com/reactnativekeyboardcontroller/traversal/ViewHierarchyNavigator.kt
@@ -5,6 +5,7 @@ import android.view.ViewGroup
 import android.widget.EditText
 import com.facebook.react.bridge.UiThreadUtil
 import com.reactnativekeyboardcontroller.extensions.focus
+import com.reactnativekeyboardcontroller.views.KeyboardToolbarExcludeReactViewGroup
 
 object ViewHierarchyNavigator {
   fun setFocusTo(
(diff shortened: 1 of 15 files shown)


kirillzyusko:
A slightly better approach would be to introduce a component:

```tsx
<KeyboardToolbar.Group name="group1" discoverable>
</KeyboardToolbar.Group>
```

In this what effectively we would do is to split inputs in group under their own names. Such approach is more flexible in my inderstanding, because if we would have tab-view written in pure JS:

|Tab1|Tab2|Tab3|
|-----|-----|----|

Then we in last input of tab 1 we could easily go to the second tab. If we would use `KeyboardToolbar.Exclude` it would serve our purposes, but the mental model here would be slightly corrupted. What effectively we want to achieve is to have groups of inputs and not "exclude" part of inputs from main view hierarchy.

I think a new API should offer better extendability for future goals.

android/src/main/java/com/reactnativekeyboardcontroller/traversal/ViewHierarchyNavigator.kt
```diff
@@ -91,7 +92,7 @@ object ViewHierarchyNavigator {
 
     if (isValidTextInput(child)) {
       result = child as EditText
-    } else if (child is ViewGroup) {
+    } else if (child is ViewGroup && child !is KeyboardToolbarExcludeReactViewGroup) {
```
IvanIhnatsiuk, line 95: this check repeats twice, would it be better to use helper function for this?

ios/traversal/ViewHierarchyNavigator.swift
```diff
@@ -91,6 +91,8 @@ public class ViewHierarchyNavigator: NSObject {
       return validTextInput
     }
 
+    guard String(describing: type(of: view)) != "KeyboardToolbarExcludeView" else { return nil }
```
IvanIhnatsiuk, line 94: 1. is it possible to use instance check instead of comparing two strings?
2. it repeats twice

PR #886, opened 2025-03-26
Update correct path of RNKCOverKeyboardViewComponentDescriptor.h in  …

### LINKED ISSUE #885: [iOS] iOS build fails.
**Describe the bug**
When i build the app on ios it doesn't build giving me 2 errors.
On example repo it doesn't happen it only happens on my main code.
I removed package-lock.json, ios folder and reinstalled the packages and ran npx expo prebuild --clean and the errors still persist.
I'll try to make a repo to replicate for now it didn't work to replicate.

**Code snippet**
```

❌  (node_modules/react-native-keyboard-controller/ios/views/OverKeyboardViewManager.mm:22:9)

  20 | #import <react/renderer/components/reactnativekeyboardcontroller/Props.h>
  21 | #import <react/renderer/components/reactnativekeyboardcontroller/RCTComponentViewHelpers.h>
> 22 | #import <reactnativekeyboardcontroller/RNKCOverKeyboardViewComponentDescriptor.h>
     |         ^ 'reactnativekeyboardcontroller/RNKCOverKeyboardViewComponentDescriptor.h' file not found
  23 | 
  24 | #import "RCTFabricComponentsPlugins.h"
  25 | #endif

› Compiling react-native-keyboard-controller Pods/react-native-keyboard-controller » KeyboardGestureAreaManager.mm

❌  (node_modules/react-native-keyboard-controller/ios/views/KeyboardGestureAreaManager.mm:20:9)

  18 | #import <react/renderer/components/reactnativekeyboardcontroller/Props.h>
  19 | #import <react/renderer/components/reactnativekeyboardcontroller/RCTComponentViewHelpers.h>
> 20 | #import <reactnativekeyboardcontroller/RNKCKeyboardGestureAreaComponentDescriptor.h>
     |         ^ 'reactnativekeyboardcontroller/RNKCKeyboardGestureAreaComponentDescriptor.h' file not found
  21 | 
  22 | #import "RCTFabricComponentsPlugins.h"
  23 | #endif
```

**Repo for reproducing**
Trying to figure out how to replicate.

**To Reproduce**

1. Build on ios with main branch
2. Doesn't build

**Expected behavior**
Should build on ios.

**Screenshots**

Image

- Desktop OS: MacOs 15.2 (24C101)
- Device: Iphone 16
- OS: latest
- RN version: 0.76.7
- RN architecture: new
- JS engine: idk
- Library version: main branch.

**Additional context**
On this [repo](https://github.com/Mako-L/problem-OverKeyboardView/) error doesn't happen on build. Will try to figure out how to replicate.

Mako-L:
…OverKeyboardViewManager.mm

Update correct path of RNKCOverKeyboardViewComponentDescriptor.h in  OverKeyboardViewManager.mm

## 📜 Description

Update correct path of RNKCOverKeyboardViewComponentDescriptor.h in  OverKeyboardViewManager.mm

## 💡 Motivation and Context

It fixes build of ios and closes #885 

## 📢 Changelog

Update correct path of RNKCOverKeyboardViewComponentDescriptor.h in  OverKeyboardViewManager.mm

### JS

-
-

### iOS

-
-

### Android

-
-

I managed to build after this change :).

## 📝 Checklist

- [X] Correct path of RNKCOverKeyboardViewComponentDescriptor.h

### DIFF
--- a/ios/views/KeyboardControllerView.mm
+++ b/ios/views/KeyboardControllerView.mm
@@ -23,7 +23,7 @@
 #import <react/renderer/components/reactnativekeyboardcontroller/EventEmitters.h>
 #import <react/renderer/components/reactnativekeyboardcontroller/Props.h>
 #import <react/renderer/components/reactnativekeyboardcontroller/RCTComponentViewHelpers.h>
-#import <reactnativekeyboardcontroller/RNKCKeyboardControllerViewComponentDescriptor.h>
+#import <react/renderer/components/reactnativekeyboardcontroller/RNKCKeyboardControllerViewComponentDescriptor.h>
 
 #import "KeyboardControllerModule-Header.h"
 #import "RCTFabricComponentsPlugins.h"
--- a/ios/views/KeyboardGestureAreaManager.mm
+++ b/ios/views/KeyboardGestureAreaManager.mm
@@ -17,7 +17,7 @@
 #import <react/renderer/components/reactnativekeyboardcontroller/EventEmitters.h>
 #import <react/renderer/components/reactnativekeyboardcontroller/Props.h>
 #import <react/renderer/components/reactnativekeyboardcontroller/RCTComponentViewHelpers.h>
-#import <reactnativekeyboardcontroller/RNKCKeyboardGestureAreaComponentDescriptor.h>
+#import <react/renderer/components/reactnativekeyboardcontroller/RNKCKeyboardGestureAreaComponentDescriptor.h>
 
 #import "RCTFabricComponentsPlugins.h"
 #endif
--- a/ios/views/OverKeyboardViewManager.mm
+++ b/ios/views/OverKeyboardViewManager.mm
@@ -19,7 +19,7 @@
 #import <react/renderer/components/reactnativekeyboardcontroller/EventEmitters.h>
 #import <react/renderer/components/reactnativekeyboardcontroller/Props.h>
 #import <react/renderer/components/reactnativekeyboardcontroller/RCTComponentViewHelpers.h>
-#import <reactnativekeyboardcontroller/RNKCOverKeyboardViewComponentDescriptor.h>
+#import <react/renderer/components/reactnativekeyboardcontroller/RNKCOverKeyboardViewComponentDescriptor.h>
 
 #import "RCTFabricComponentsPlugins.h"
 #endif
--- a/react-native-keyboard-controller.podspec
+++ b/react-native-keyboard-controller.podspec
@@ -61,7 +61,7 @@ Pod::Spec.new do |s|
   if new_arch_enabled
     s.subspec "common" do |ss|
       ss.source_files         = ["common/cpp/**/*.{cpp,h}"]
-      ss.header_dir           = "reactnativekeyboardcontroller"
+      ss.header_dir           = "react/renderer/components/reactnativekeyboardcontroller"
       ss.private_header_files = "common/cpp/**/*.{h}"
       ss.pod_target_xcconfig  = { "HEADER_SEARCH_PATHS" => "\"$(PODS_TARGET_SRCROOT)/common/cpp\"" }
     end


kirillzyusko:
Hey @Mako-L 

Thanks for the PR. I've got two questions:
- why you changed only a single import? Would it make sense to change all other imports (`KeyboardGestureArea`, `KeyboardControllerView`)
- does it make sense to change `ss.header_dir = "reactnativekeyboardcontroller"` to `ss.header_dir = "react/renderer/components/reactnativekeyboardcontroller"` as well in this PR?

kirillzyusko:
Also don't know how CI passed, but locally I'm getting this error now 🙃 

image

Mako-L:
@kirillzyusko In the image i see it still has bad path maybe you modified it look at the upper imports.

Mako-L:
Closed it by mistake.

Mako-L:
@kirillzyusko ```#import <react/renderer/components/reactnativekeyboardcontroller/EventEmitters.h>
#import <react/renderer/components/reactnativekeyboardcontroller/Props.h>
#import <react/renderer/components/reactnativekeyboardcontroller/RCTComponentViewHelpers.h>
#import <react/renderer/components/reactnativekeyboardcontroller/RNKCOverKeyboardViewComponentDescriptor.h>``` Look at the upper imports they all need to be the same to import from `react/renderer/components/reactnativekeyboardcontroller/`

kirillzyusko:
> In the image i see it still has bad path maybe you modified it look at the upper imports.

That's right, I haven't woken up yet this morning 😅

kirillzyusko:
> I only changed that one because it was the only one that had wrong import path reactnativekeyboardcontroller instead of react/renderer/components/reactnativekeyboardcontroller/ like the upper imports and it was the only line needed to change to fix the ios build.

But other components also have a different path, for example:

```objc
#import <reactnativekeyboardcontroller/RNKCKeyboardGestureAreaComponentDescriptor.h>
#import <reactnativekeyboardcontroller/RNKCKeyboardControllerViewComponentDescriptor.h>
```

Interesting why they didn't throw compilation errors for you? What happens if you press cmd+click on them - will you be redirected to `common` folder?..

kirillzyusko:
I'm not against to change the path, but just want to have consistency in all files and in Pod declaration too 🙂

Mako-L:
Thing is modifying this path makes it consistent with the upper imports. If we could make it see `reactnativekeyboardcontroller/` path work instead of `react/renderer/components/reactnativekeyboardcontroller` the upper imports should be modified as well to keep consistency. But as you can see in this file consistancy means to use `react/renderer/components/reactnativekeyboardcontroller`. I don't have the right extension in vscode to render the .mm file corectly so I can't click on the path 😄

Mako-L:
I see this path is used in more files there is the imports 3 imports from `react/renderer/components/reactnativekeyboardcontroller` and one `reactnativekeyboardcontroller`. I'm not sure how this works correctly. 
image

Mako-L:
So from what i understand the problem is that file we trying to import is generated and not from module's own header files directly.

kirillzyusko:
Those files were previously auto-generated, and to stretch OverKeyboardView to full screen on Android I had to copy them and manually modify.
So now they are not auto-generated. And I thought to move them under separate header to make it clearer that they are not autogenerated.
However similar approach was used in `react-native-safe-area-context`. There also custom shadow nodes are used, but the import is under `react/renderer/components` path.

So my idea is to rename two other files (KeyboardGestureArea and KeyboardControllerView) and change the `header_dir` in Podfile 🙃 

If CI pass then we can merge it 😎

kirillzyusko:
@Mako-L seems like CI passed, can you check new changes in your project? Do they will resolve compilation errors?

Mako-L:
@kirillzyusko It seems to work fine and built the ios build ok. Sorry i get caught in other stuff and forgot to recheck until now.

PR #894, opened 2025-04-01
feat: useKeyboardState

kirillzyusko:
## 💡 Motivation and Context

Closes https://github.com/kirillzyusko/react-native-keyboard-controller/issues/893

### JS

-
-

### iOS

-
-

### Android

-
-

## 📝 Checklist

- [ ] CI successfully passed

### DIFF
--- a/src/hooks/useKeyboardState/index.ts
+++ b/src/hooks/useKeyboardState/index.ts
@@ -1,0 +1,34 @@
+import { useEffect, useState } from "react";
+
+import { KeyboardEvents } from "../../bindings";
+import { KeyboardController } from "../../module";
+
+// TODO: expose isVisible, isTransitioning (isHiding/isShowing) props?
+export const useKeyboardState = () => {
+  const [state, setState] = useState(KeyboardController.state());
+
+  useEffect(() => {
+    const didShowSubscription = KeyboardEvents.addListener(
+      "keyboardDidShow",
+      (e) => {
+        setState(e);
+      },
+    );
+    const didHideSubscription = KeyboardEvents.addListener(
+      "keyboardDidHide",
+      (e) => setState(e),
+    );
+
+    // we might have missed an update between reading a value in render and
+    // `addListener` in this handler, so we set it here. If there was
+    // no change, React will filter out this update as a no-op.
+    setState(KeyboardController.state());
+
+    return () => {
+      didShowSubscription.remove();
+      didHideSubscription.remove();
+    };
+  }, []);
+
+  return state;
+};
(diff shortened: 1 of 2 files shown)


src/hooks/useKeyboardState/index.ts
```diff
@@ -0,0 +1,34 @@
+import { useEffect, useState } from "react";
+
+import { KeyboardEvents } from "../../bindings";
+import { KeyboardController } from "../../module";
+
+// TODO: expose isVisible, isTransitioning (isHiding/isShowing) props?
+export const useKeyboardState = () => {
+  const [state, setState] = useState(KeyboardController.state());
```
IvanIhnatsiuk, line 8: should it be?

```suggestion
  const [state, setState] = useState(() => KeyboardController.state());
```

src/hooks/useKeyboardState/index.ts
```diff
@@ -0,0 +1,34 @@
...
+  useEffect(() => {
+    const didShowSubscription = KeyboardEvents.addListener(
+      "keyboardDidShow",
+      (e) => {
+        setState(e);
+      },
+    );
+    const didHideSubscription = KeyboardEvents.addListener(
+      "keyboardDidHide",
+      (e) => setState(e),
```
IvanIhnatsiuk, line 19: you don't need to create a new function for listener

```suggestion
      setState,
```

src/hooks/useKeyboardState/index.ts
```diff
@@ -0,0 +1,34 @@
...
+// TODO: expose isVisible, isTransitioning (isHiding/isShowing) props?
+export const useKeyboardState = () => {
+  const [state, setState] = useState(KeyboardController.state());
+
+  useEffect(() => {
+    const didShowSubscription = KeyboardEvents.addListener(
+      "keyboardDidShow",
+      (e) => {
+        setState(e);
+      },
```
IvanIhnatsiuk, line 15: you don't need to create a new function for listener

```suggestion
     setState
```

src/hooks/useKeyboardState/index.ts
```diff
@@ -0,0 +1,34 @@
...
+    );
+    const didHideSubscription = KeyboardEvents.addListener(
+      "keyboardDidHide",
+      (e) => setState(e),
+    );
```
IvanIhnatsiuk, line 20: so, you can create an array with handled events outside of hook.

```
const EVENTS = ['keyboardDidShow', 'keyboardDidHide'];
```

then in useEffect you can use it like that:

```
useEffect(() => {
   const subscribers = keyboardDidHide.map((event) => KeyboardEvents.addListener(event, setState))

   return () => subscribers.forEach((subscriber) => subscriber.remove())
}, [])
```

PR #901, opened 2025-04-03
 [W.I.P] KeyboardAwareScrollView Rework To work with multilines

Bowlerr:
## 📜 Description

Will write up a proper description later and clean up the code.

current issue until done (Android):

- The multiple line inputs always start the text cursor from the bottom on first focus
-  - ideally I want it to start where I press

- 

### JS

-
-

### iOS

-
-

### Android

-
-

## 📝 Checklist

- [ ] CI successfully passed

### DIFF
--- a/example/src/screens/Examples/AwareScrollView/index.tsx
+++ b/example/src/screens/Examples/AwareScrollView/index.tsx
@@ -13,10 +13,6 @@ import type { StackScreenProps } from "@react-navigation/stack";
 type Props = StackScreenProps<ExamplesStackParamList>;
 const snapToOffsets = [125, 225, 325, 425, 525, 625];
 
-const BIG_TEXT_TWO = `Where does it come from?
-Contrary to popular belief, Lorem Ipsum is not simply random text. It has roots in a piece of classical Latin literature from 45 BC, making it over 2000 years old. Richard, a Latin professor at Hampden-Sydney College in Virginia, looked up one of the more obscure Latin words, consectetur, from a Lorem Ipsum passage, and going through the cites of the word in classical literature, discovered the certain source. Lorem Ipsum comes from sections 1.10.32 and 1.10.33 of The Extremes of Good and Evil by Cicero, written in 45 BC. This book is a treatise on the theory of ethics, very popular during the Renaissance. The first line of Lorem Ipsum, "Lorem ipsum dolor sit..", comes from a line in section 1.10.32.
-`;
-
 const BIG_TEXT = `Where does it come from?
 Contrary to popular belief, Lorem Ipsum is not simply random text. It has roots in a piece of classical Latin literature from 45 BC, making it over 2000 years old. Richard, a Latin professor at Hampden-Sydney College in Virginia, looked up one of the more obscure Latin words, consectetur, from a Lorem Ipsum passage, and going through the cites of the word in classical literature, discovered the certain source. Lorem Ipsum comes from sections 1.10.32 and 1.10.33 of The Extremes of Good and Evil by Cicero, written in 45 BC. This book is a treatise on the theory of ethics, very popular during the Renaissance. The first line of Lorem Ipsum, "Lorem ipsum dolor sit..", comes from a line in section 1.10.32.
 
@@ -93,9 +89,10 @@ export default function AwareScrollView({ navigation }: Props) {
           </>
         )}
 
-        {new Array(5).fill(0).map((_, i) => (
+        {new Array(10).fill(0).map((_, i) => (
           <TextInput
             key={i}
+            contextMenuHidden={i === 4 && Platform.OS === "ios"}
             keyboardType={i % 2 === 0 ? "numeric" : "default"}
             placeholder={`TextInput#${i}`}
             onChangeText={setText}
@@ -103,21 +100,6 @@ export default function AwareScrollView({ navigation }: Props) {
         ))}
         <TextInput
           defaultValue={BIG_TEXT}
-          placeholder="TextInput#4"
-          style={styles.input}
-          onChangeText={setText}
-        />
-        {new Array(5).fill(0).map((_, i) => (
-          <TextInput
-            key={i + 5}
-            contextMenuHidden={i === 4 && Platform.OS === "ios"}
-            keyboardType={i % 2 === 0 ? "numeric" : "default"}
-            placeholder={`TextInput#${i + 5}`}
-            onChangeText={setText}
-          />
-        ))}
-        <TextInput
-          defaultValue={BIG_TEXT_TWO}
           placeholder="TextInput#10"
           style={styles.input}
           onChangeText={setText}
--- a/example/src/screens/Examples/AwareScrollView/index.tsx
+++ b/example/src/screens/Examples/AwareScrollView/index.tsx
@@ -13,6 +13,10 @@ import type { StackScreenProps } from "@react-navigation/stack";
 type Props = StackScreenProps<ExamplesStackParamList>;
 const snapToOffsets = [125, 225, 325, 425, 525, 625];
 
+const BIG_TEXT_TWO = `Where does it come from?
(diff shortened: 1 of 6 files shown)


Bowlerr:
https://github.com/user-attachments/assets/3b2d245e-edf2-4211-b3f5-6da3d334faf6

Bowlerr:
@kirillzyusko lmk what you think

src/components/KeyboardAwareScrollView/index.tsx
```diff
@@ -345,42 +586,63 @@ const KeyboardAwareScrollView = forwardRef<
...
+    );
+
     return (
       <ScrollViewComponent
         ref={onRef}
         {...rest}
         scrollEventThrottle={16}
+        onContentSizeChange={onContentSizeChange}
         onLayout={onScrollViewLayout}
+        onScroll={scrollHandler}
```
kirillzyusko, line 645: It's not very trivial to pass handler in such way. We have to forward that event to external handlers. And it's tricky, because those handlers can be worklet/plain-js handlers. I'll link a chain of PR here in chronological order: https://github.com/kirillzyusko/react-native-keyboard-controller/pull/339 https://github.com/kirillzyusko/react-native-keyboard-controller/pull/408 https://github.com/kirillzyusko/react-native-keyboard-controller/pull/490 https://github.com/kirillzyusko/react-native-keyboard-controller/pull/500

src/components/KeyboardAwareScrollView/index.tsx
```diff
@@ -345,42 +586,63 @@ const KeyboardAwareScrollView = forwardRef<
...
+        paddingBottom: enabled ? currentKeyboardFrameHeight.value : 0,
+      }),
       [enabled],
     );
 
+    useAnimatedReaction(
+      () => scrollY.value,
+      (current) => {
+        scrollTo(scrollViewAnimatedRef, 0, current, false);
+      },
```
kirillzyusko, line 635: I think it gives an effect of "keep strong distance between caret and keyboard"

I think default iOS/Android implementation perform scroll only if needed, i. e. here

image

When you move pointer higher:

image

It scrolls (but the expected behavior is that it shouldn't)

src/components/KeyboardAwareScrollView/index.tsx
```diff
@@ -95,25 +109,167 @@ const KeyboardAwareScrollView = forwardRef<
...
+              "EndingPosition:",
+              scrollY.value,
+              "Distance From Goal:",
+              Math.abs(scrollOffsetY.value - goal),
+              "Finished:",
+              finished,
+            );
+            isScrolling.value = false;
+          });
+    };
```
kirillzyusko, line 272: Why do we need this? We should perform scroll based on a keyboard position? So we should scroll from `onMove` handlers.

Plain scroll can happen when text input grows, but in this case `scrollTo(..., true)` with default animation works also pretty well 🤔

example/src/screens/Examples/AwareScrollView/index.tsx
```diff
@@ -89,17 +93,31 @@ export default function AwareScrollView({ navigation }: Props) {
           </>
         )}
 
-        {new Array(10).fill(0).map((_, i) => (
+        {new Array(5).fill(0).map((_, i) => (
```
kirillzyusko, line 96: If you modify code example in such way there is 100% guarantee that e2e tests will fail, because they expect to have certain amount of inputs.

I'd suggest to revert these changes (I understand that it's more convenient not scroll through entire form, but you can keep these changes locally and not commit them to the repository)

example/src/screens/Examples/AwareScrollView/index.tsx
```diff
@@ -13,6 +13,10 @@ import type { StackScreenProps } from "@react-navigation/stack";
 type Props = StackScreenProps<ExamplesStackParamList>;
 const snapToOffsets = [125, 225, 325, 425, 525, 625];
 
+const BIG_TEXT_TWO = `Where does it come from?
```
kirillzyusko, line 16: Maybe let's not repeat code and re-use variables?

But again, for playground code you can keep a local changes and not commit them to the repository.

example/src/screens/Examples/AwareScrollView/index.tsx
Bowlerr: Planned to clean this up and wasn't suppose to commit the changes to the example.

example/src/screens/Examples/AwareScrollView/index.tsx
Bowlerr: Planned to clean this up and wasn't suppose to commit the changes to the example.

src/components/KeyboardAwareScrollView/index.tsx
Bowlerr: I can change this. I liked the effect for inputs that don't fit inside the visible area. Without this the user would need to scroll themselves or beginning to type for the move to happen. 

I can change it so the scroll stops at the top of the input instead? and then follows from the caret at bottom until it reaches the bottom of the input where it will sit if it has a maxHeight?

kirillzyusko:
And feedback based on video:
- 0:08-0:09 it feels a little bit wrong, that you see scroll and at certain point of time this scroll continues. Do we have selection coordinates before `onStart`-keyboard event? If not then it can be a bug in keyboard-controller and I think it should be fixed (basically before `onStart` you should have these coordinates, because we already have focused input)
- 0:13-0:15 - I think we shouldn't scroll to follow cursor position. If scroll remains in a visible area then better not scroll, since user can still see the input.
- 0:25-0:27 - feels like scroll is stuttering a little bit?
- 0:50-0:51 - why we scrolled such a big distance? I thought we always rely on selection coordinates (not on layout anymore)?
- 0:59-1:00 - animation feels desynchronized?
- when back scroll turned off it looks like you add back scroll? And when it's enabled I don't see a back scroll?
- not sure, but since you modified example it looks like snap points are broken now

Overall you did a great job 💪  But certain points must be addressed I don't think I can merge PR in a current state 🤔 

But if you don't want to spend too much time on going back-and-forth with review, then you can use that modified version of `KeyboardAwareScrollView` in your project, since that component uses public API of the lib 🙂 

Also if you want we can have a conversation in discord, it should significantly improve our communication process 😊

Bowlerr:
I'm going away this weekend so have not much time but would love to jump on discord next week if that works with you @kirillzyusko 

And feedback based on video:

**0:08-0:09 - feels off**
Agreed, I believe this is because of the withSpring. Will remove and find another solution to the issue it solved. 
the issue I was solving was: on iOS when you select the bottom component, it is off by a few pixels ( if you remove the margin from the text input). My solution is hacky for such an edge case so happy to remove it.

**0:13-0:15 - I think we shouldn't scroll to follow cursor position. If scroll remains in a visible area then better not scroll, since user can still see the input.**
- Can change to this, can always add a prop to follow cursor on long inputs?

**0:25-0:27 - feels like scroll is stuttering a little bit?**
- need to work on animation timings or could be slow in debug mode on a simulator which i've experienced a lot in the past.

**0:50-0:51 - why we scrolled such a big distance? I thought we always rely on selection coordinates (not on layout anymore)?**
- If the input fit within the viewable space in the scrollview -> Scroll input to fill view
- if it doesn't -> follow selection coordinates

**0:59-1:00 - animation feels desynchronized?**
- agree, I'll look into this.

**when back scroll turned off it looks like you add back scroll? And when it's enabled I don't see a back scroll?**
- On the example, when toggle is enabled, it disables the back scroll. its enabling the disabling.

**not sure, but since you modified example it looks like snap points are broken now**
- How are they supposed to work? 
- It looks correct to me as they are snapping the points to the top of the scroll?

Bowlerr:
> you can use that modified version of KeyboardAwareScrollView in your project, since that component uses public API of the lib 🙂

Already on it 😎. Thanks for the great api

kirillzyusko:
I'll close this PR in favour of https://github.com/kirillzyusko/react-native-keyboard-controller/pull/1026

It'll be published under `1.18.0` hopefully next weel!

PR #912, opened 2025-04-10
fix: findNodeHandle on the web platform

IvanIhnatsiuk:
## 📜 Description

In the new version of `react-native-web` they added an error throw if you try to use findNodeHandle from `react-native` https://github.com/necolas/react-native-web/blob/922c134f2b7c428cc19daaeb3cac4b6a4c8ec6a3/packages/react-native-web/src/exports/findNodeHandle/index.js#L11. So I created a findNodeHandle utility that uses react-native's findNodeHandle for the native platform and returns Node or null for web.

## 💡 Motivation and Context

This library should work well on all platforms

### JS

- added findNodeHandle util

tested in a web browser

Before:

https://github.com/user-attachments/assets/7056e962-3427-4127-86ad-c204d09df874

After: 

https://github.com/user-attachments/assets/a2593a86-38dd-4f8d-aa73-0be04f1b3227

## 📝 Checklist

- [x] CI successfully passed

### DIFF
--- a/src/components/KeyboardAwareScrollView/index.tsx
+++ b/src/components/KeyboardAwareScrollView/index.tsx
@@ -1,5 +1,4 @@
 import React, { forwardRef, useCallback, useMemo } from "react";
-import { findNodeHandle } from "react-native";
 import Reanimated, {
   interpolate,
   scrollTo,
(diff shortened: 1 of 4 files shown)


kirillzyusko:
Since I don't support web - it's okay to return element as node-handle.

On web for `registerEventHandler` I use mock, `KeyboardAwareScrollView` also will work with these changes (it will compare node-handle with `null`, so will be `false` and will not work on web (and it's expected)).

PR #956, opened 2025-05-24
Add width and height as optional parameters for OverKeyboardView 

### LINKED ISSUE #955: new useWindowDimensions on ios has values swapped
**Describe the bug**
The new useWindowDimensions function we added on ios seems to have lag or something it  takes dimensions on rotation the opposite like hight is width and width is height i tested on Ipad 11 and on device and just saw the issue. Seems to be this issue https://github.com/facebook/react-native/issues/29290 / https://github.com/facebook/react-native/issues/49511. Is weird cause when tested on the repo that I shared in previous bug it was working at first glance but on my main repo it seems values are swapped even with screen approach. I used tablet this time and seems the bug is there. I will update more this issue when I get more time on my hand but for now I just wanted to let you know about this.

**Code snippet**
Add your code snippet where error has been occurred.

**Repo for reproducing**
https://github.com/Mako-L/problem-OverKeyboardView

**To Reproduce**

1. Open the app on Ipad 
2. Go to chat
3. rotate screen
4. click attach icon button
5. see problem

**Expected behavior**
It should take the dimensions correctly and have full screen modal but dimensions are swapped due to rn bug on Dimensions.

**Screenshots**
Image

- Desktop OS: Mac latest
- Device: Ipad 11
- OS: 18.2
- RN version: 0.79.1 and 0.79.2
- RN architecture: new
- JS engine: idk
- Library version: github:kirillzyusko/react-native-keyboard-controller#main

**Additional context**
I think we should send the width and height from outside the module and use useSafeAreaFrame() to send them. I think we should add them as optional parameters and use if not null what do you think?

Mako-L:
## 📜 Description

I added width and height as optional parameters for OverKeyboardView to be added externally for people that have the issue that useDimensions hook doesn't work properly on Ipad or other ios devices. Closes #955 
## 💡 Motivation and Context

useDimensions is not stable for ios devices.
People should be able to give external width and height values to OverKeyboardView.

## 📢 Changelog

added width and height as optional parameters on OverKeyboardView

### JS

added width and height as optional parameters on OverKeyboardView

### iOS

-
-

### Android

-
-

I tested on Ipad and Iphone and everything seems fixed.

image

## 📝 Checklist

- [ ] CI successfully passed


src/types/views.ts
```diff
@@ -43,4 +43,6 @@ export type OverKeyboardViewProps = PropsWithChildren<{
    * A boolean prop indicating whether the view is visible or not. If it's true then view is shown on the screen. If it's false then view is hidden.
    */
   visible: boolean;
+  height?: number;
+  width?: number;
```
kirillzyusko, line 47: I'd avoid passing style-props as separate props 🤔 

Maybe we should add optional `style` property? It would be a little bit more universal 👀

src/views/OverKeyboardView/index.tsx
```diff
@@ -4,7 +4,7 @@ import { Platform, StyleSheet, View } from "react-native";
 import { RCTOverKeyboardView } from "../../bindings";
 import { useWindowDimensions } from "../../hooks";
 
-import type { OverKeyboardViewProps } from "../../types";
+import type { OverKeyboardViewProps } from "../../types/views";
```
kirillzyusko, line 7: Why you changed import here?

src/views/OverKeyboardView/index.tsx
```diff
@@ -20,10 +20,20 @@ import type { PropsWithChildren } from "react";
...
-const OverKeyboardView = (props: PropsWithChildren<OverKeyboardViewProps>) => {
-  const { children, visible } = props;
-  const { height, width } = useWindowDimensions();
-  const inner = useMemo(() => ({ height, width }), [height, width]);
+const OverKeyboardView = ({
+  children,
+  visible,
+  height: propHeight,
+  width: propWidth,
+}: PropsWithChildren<OverKeyboardViewProps>) => {
```
kirillzyusko, line 28: Let's not use destruction here, because jsdoc plugin will require to write a lot of jsdoc 😅

kirillzyusko (changes requested):
As a temporary solution I think it's okay to use this approach, but maybe let's not pass `width`/`height` as props and let's simply pass generic `style` property?

In you code you can specify your own width/height with a comment why you do that, and in a meantime I promise to look into that problem 👀

Issue #974, opened 2025-06-11
iOS 26: `KeyboardToolbar` doesn't match new UI guidelines


kirillzyusko:
**Describe the bug**

`KeyboardToolbar` doesn't match new UI guidelines on iOS 26.

**Code snippet**

Example app.

**Repo for reproducing**

Example app.

**To Reproduce**

1. Go to KeyboardToolbar example
2. Click on Input
3. Check how it looks

**Expected behavior**

Well, at least according to a new design there:
- should be an empty space between keyboard and toolbar;
- should have rounded corners;
- should have ✔ icon instead of done;
- should have glass effect;

**Screenshots**

|Expected (v1)|Expected (v2)|Actual|IQKeyboardManager|
|--------------|--------------|------|--------------------|
|<img width="150" src="https://github.com/user-attachments/assets/6e42fa48-f98d-43fa-8d85-a17c5c4947c8">|<img width="150" src="https://github.com/user-attachments/assets/3b71a61a-0765-4598-bb5b-4cd0018f3888">|<img width="150" src="https://github.com/user-attachments/assets/f0b4787f-c08f-4750-b54a-329e95ec08be">|<img width="150" src="https://github.com/user-attachments/assets/6500200d-c8a0-444d-b522-de72c4d6bb6d">|

**Smartphone (please complete the following information):**

- Desktop OS: macOS 26
- Device: iPhone 16
- OS: iOS 26
- RN version: 0.79.0
- RN architecture: old
- JS engine: Hermes
- Library version: 1.17.4

**Additional context**

Let's check IQKeyboardManager implementation to see how it works there 👀

PR #978, opened 2025-06-12
fix: `KeyboardToolbar` on iOS 26

kirillzyusko:
## 📜 Description

Adopt `KeyboardToolbar` UI for iOS 26.

## 💡 Motivation and Context

⚠️ This is a temporary solution. ⚠️ 

Starting from iOS 26 we:
- have rounder corners for keyboard;
- proposed to use liquid glass effect.

An original implementation from Safari now uses rounded corners + glass effect. Also instead of "Done" button Apple shows ✔️ Overall current Safari implementation doesn't look so well (in my opinion).

If we have a look how `IQKeyboardManager` handles that... well, it's even worse:

<p align="center">

</p>

So I think the UI for this element is not finalized yet. 

Anyway current implementation in keyboard-controller doesn't look well at all, so in this PR I'm:
- rounding corners;
- adding horizontal margins (so toolbar doesn't take full width);
- moving toolbar a little bit up, so that it has a small spacing between floating element and keyboard.

This UI is "okay" in my understanding (but still not perfect). As a temporary solution I think it's good enough, however I'm currently in a big phase of research with `KeyboardExtender` and `KeyboardBackgroundView`. Theoretically we may want to utilize these components in the future or we can continue our research to see what is the new way of building keyboard "extenders" and how it can be back-ported to older versions.

Closes https://github.com/kirillzyusko/react-native-keyboard-controller/issues/974

### JS

- move `KeyboardToolbar` constant to separate file;
- add `KEYBOARD_HAS_ROUNDED_CORNERS` variable;
- make toolbar "floating" in case if keyboard has rounded corners;

Tested manually on iPhone 16 Pro (iOS 26).

|Safari|Keyboard Controller|
|------|--------------------|
|||

## 📝 Checklist

- [x] CI successfully passed

### DIFF
--- a/src/components/KeyboardToolbar/index.tsx
+++ b/src/components/KeyboardToolbar/index.tsx
@@ -1,5 +1,5 @@
 import React, { useCallback, useEffect, useMemo, useState } from "react";
-import { Platform, StyleSheet, Text, View } from "react-native";
+import { StyleSheet, Text, View } from "react-native";
 
 import { FocusedInputEvents } from "../../bindings";
 import { KeyboardController } from "../../module";
@@ -9,6 +9,17 @@ import KeyboardStickyView from "../KeyboardStickyView";
 import Arrow from "./Arrow";
 import Button from "./Button";
 import { colors } from "./colors";
+import {
+  DEFAULT_OPACITY,
+  KEYBOARD_HAS_ROUNDED_CORNERS,
+  KEYBOARD_TOOLBAR_HEIGHT,
+  OPENED_OFFSET,
+  TEST_ID_KEYBOARD_TOOLBAR,
+  TEST_ID_KEYBOARD_TOOLBAR_CONTENT,
+  TEST_ID_KEYBOARD_TOOLBAR_DONE,
+  TEST_ID_KEYBOARD_TOOLBAR_NEXT,
+  TEST_ID_KEYBOARD_TOOLBAR_PREVIOUS,
+} from "./constants";
 
 import type { HEX, KeyboardToolbarTheme } from "./types";
 import type { KeyboardStickyViewProps } from "../KeyboardStickyView";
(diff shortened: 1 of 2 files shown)


src/components/KeyboardToolbar/index.tsx
```diff
@@ -73,6 +73,9 @@ const TEST_ID_KEYBOARD_TOOLBAR_DONE = `${TEST_ID_KEYBOARD_TOOLBAR}.done`;
 
 const KEYBOARD_TOOLBAR_HEIGHT = 42;
 const DEFAULT_OPACITY: HEX = "FF";
+const KEYBOARD_HAS_ROUNDED_CORNERS =
+  Platform.OS === "ios" && parseInt(Platform.Version, 10) >= 26;
+const OPENED_OFFSET = KEYBOARD_HAS_ROUNDED_CORNERS ? -10 : 0;
```
IvanIhnatsiuk, line 78: would it be better to move all of these constants to a separate folder?

IvanIhnatsiuk, line 78: could you tell me how did you measure that it should be -10?

IvanIhnatsiuk: Hmm, borders are different. Looks like apple uses squircle for this element. Have you tried to use borderCurve style prop?

src/components/KeyboardToolbar/index.tsx
```diff
@@ -258,6 +265,13 @@ const styles = StyleSheet.create({
     marginRight: 16,
     marginLeft: 8,
   },
+  floating: {
+    width: "92%",
```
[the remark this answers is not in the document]
IvanIhnatsiuk: Okay, and will it be the same on iPads in landscape/portrait modes?

PR #981, opened 2025-06-14
feat: `KeyboardBackgroundView`

kirillzyusko:
## 📜 Description

Added `KeyboardBackgroundView` - component designed to match keyboard backdrop UI.

## 💡 Motivation and Context

This component can be useful when you want to continue keyboard surface and embed advanced elements (such as `TextInput` inside it). Using this component you can achieve a similar UI to Photos app or Safari app search on iOS. Additioanlly this component will be used as a polyfill implementation for `KeyboardExtender` on Android (just because `KeyboardExtender` can not be implemented natively on Android).

> [!WARNING]
> This component only matches default keyboard UI background. If user customizes it -> it will look differently and will not match keyboard design. There is no way to check displayed keyboard color on a native side, but some apps still use this approach, so I thought to export that possibility from `keyboard-controller` too.

On iOS we can use system class `UIKBBackdropView` to match the keyboard background. On Android keyboard is a 3rd party process so there is no API at all, so I kind of reverse-engineered it: we have an access to keyboard package name and then based on pacakge name and phone preferences we can "guess" keyboard color. In reality it may vary from OS version o OS version and may be even dynamic and depend on wallpaper generic color. To solve it I Added a class that has a basic color (can be dynamic too) and then we have additional property called `tone`, which can slightly adjust the color. The idea is that color like `#3c3c3c` with tone `+2` will become `#3e3e3e`. For now it works and we can successfully match default keyboard color.

### E2E

- cover `KeyboardBackgroundView` by e2e tests;

### Docs

- mention new feature in README;
- new docs page `KeyboardBackgroundView`;
- new chapter in Components Overview guide.

### JS

- added `KeyboardBackgroundView`;
- added unit-tests for `KeyboardBackgroundView`.

### iOS

- added `KeyboardBackgroundView`;

### Android

- added `KeyboardBackgroundView`;
- move `isSystemDarkMode` extensions to `Context`;
- create new `currentImePackage` extension;
- create resource files;

Tested manually on:
- Pixel 7 Pro (API 35, API 36 - real device);
- Pixel 9 Pro (API 31, API 33, API 34, API 35, API 36 - emulator);
- Xiaomi Redmi Note 5 Pro (API 28 - real device);
- iPhone 11 (iOS 18.5 - real device);
- iPhone 6s (iOS 15.8 - real device);
- iPhone 14 Pro (iOS 18.4 - real device).
- iPhone 16 Pro (iOS 18.5, 26.0 - simulator)
- iPhone 15 Pro (iOS 17.5 - simulator)

plus tested on all e2e devices (via e2e tests).

### Shared surface

|Android|iOS|
|--------|---|
|android|android|

### "Meatball"/Ask AI/Liquid keyboard

|Android|iOS|
|--------|---|
|<video src="https://github.com/user-attachments/assets/ac3efba6-2a6a-43a1-b0ee-501e2b5be361">|<video src="https://github.com/user-attachments/assets/2f7d5c3e-4e32-43d9-b2d4-9658accc757a">|

## 📝 Checklist

- [x] CI successfully passed


android/src/main/java/com/reactnativekeyboardcontroller/managers/KeyboardBackgroundViewManagerImpl.kt
```diff
@@ -0,0 +1,17 @@
+package com.reactnativekeyboardcontroller.managers
+
+import com.facebook.react.bridge.ReactApplicationContext
+import com.facebook.react.uimanager.ThemedReactContext
+import com.reactnativekeyboardcontroller.views.background.KeyboardBackgroundViewGroup
+
+@Suppress("detekt:UnusedPrivateProperty")
```
kirillzyusko, line 7: Should be removed after rebase

docs/docs/api/keyboard-background-view/keyboard-background-view.lottie.json
kirillzyusko, line 1: Let's re-work the animation - let's remove touch interactions, because they distract a lot 😓

example/ios/KeyboardControllerExample.xcodeproj/project.pbxproj
```diff
@@ -244,7 +244,7 @@
 						TestTargetID = 13B07F861A680F5B00A75B9A;
 					};
 					13B07F861A680F5B00A75B9A = {
-						DevelopmentTeam = LZ65PSZ7CS;
+						DevelopmentTeam = 7JPW5W7JHV;
```
kirillzyusko, line 247: Let's revert it

kirillzyusko:
This is a list of to do which will be addressed later on:

- resolve all todos in the code
- iOS 26 - doesn't match background
- re-work lottie animation
- remove hardcoded constants (like 291) from example

PR #1027, opened 2025-07-19
fix: delay delegate substitution

IvanIhnatsiuk:
## 📜 Description

Currently in my library, I have a bug that if we have text input with autofocus, your library adds a delegate and it does it before my library, as I do it with decorator view, which is mounted on the screen by the next element. I think it is not so important for your library to put its delegate( in terms of as early as possible). So I added logic that does this on the next frame.

Advanced input mask issue: https://github.com/IvanIhnatsiuk/react-native-advanced-input-mask/issues/128

## 💡 Motivation and Context

Allow other libraries to set their own delegates

### iOS

- Wrapped substituteDelegate call with DispatchQueue.main.async

tested locally on iOS simulator

https://github.com/user-attachments/assets/bb9e4725-d7d4-4fcb-9e28-0d94282315c0

## 📝 Checklist

- [x] CI successfully passed

### DIFF
--- a/ios/observers/FocusedInputObserver.swift
+++ b/ios/observers/FocusedInputObserver.swift
@@ -140,6 +140,11 @@ public class FocusedInputObserver: NSObject {
     currentInput = currentResponder?.superview as UIView?
 
     setupObservers()
+    // dispatch onSelectionChange on focus
+    if let textInput = responder as? UITextInput {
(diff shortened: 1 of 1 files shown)


kirillzyusko (approved):

kirillzyusko:
@IvanIhnatsiuk can you mention in PR description exact issue that you are trying to fix from `advanced-input-mask` repository?

PR #1047, opened 2025-07-25
feat: keyboard tracking view

kirillzyusko:
### JS

-
-

### iOS

-
-

### Android

-
-

## 📝 Checklist

- [ ] CI successfully passed


kirillzyusko:
After thinking a while I think it'll be a good idea to merge it now.

`KeyboardTrackingView` is a great abstraction that helps to isolate certain pieces of code and provides a better interface segregation. So it's anyway something that I'd like to have in the codebase 👀

PR #1059, opened 2025-07-31
fix: iOS crash caused by the `KeyboardControllerView` class implementation

zhouzh1:
## 📜 Description

We caught an iOS crash by the Sentry error monitoring, according to the stack trace, finally we found there is a bug in the `KeyboardControllerView` class, the stack trace is like the below screenshot.

image

## 💡 Motivation and Context

This PR solved this crash with a very simple fix, seems that this bug is caused by a unconscious typo error...

### iOS

- Changed the `KeyboardControllerView` class implementation to resolve this crash.

Actually this crash is occasional and at least I didn't manage to reproduce it locally, the current fix is totally from logic inference, however we have applied the fix in this PR in our prod app more than one month by patching the library file, and we didn't find any new relevant crash occurrences since the fix, so it proves to work on our end.

## 📝 Checklist

- [ ] CI successfully passed

### DIFF
--- a/ios/views/KeyboardControllerView.mm
+++ b/ios/views/KeyboardControllerView.mm
@@ -178,16 +178,16 @@ using namespace facebook::react;
                           .duration = [duration intValue],
                           .target = [target intValue]});
             }
-          }
-          if ([event isEqualToString:@"onKeyboardMoveInteractive"]) {
-            std::dynamic_pointer_cast<const facebook::react::KeyboardControllerViewEventEmitter>(
-                self->_eventEmitter)
-                ->onKeyboardMoveInteractive(
-                    facebook::react::KeyboardControllerViewEventEmitter::OnKeyboardMoveInteractive{
-                        .height = [height doubleValue],
-                        .progress = [progress doubleValue],
-                        .duration = [duration intValue],
-                        .target = [target intValue]});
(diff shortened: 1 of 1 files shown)


kirillzyusko:
Thank you! I just saw this crash few days ago in Google Crashlytics - glad you managed to find a proper fix ❤️ 

Let's wait for CI to pass and I'll merge it!

kirillzyusko:
`objc-lint` fails 🙈 

Can you remove `-n` from `xargs clang-format -i -n --Werror` and run `lint-clang`? And commit new changes?

zhouzh1:
> `objc-lint` fails 🙈
> 
> Can you remove `-n` from `xargs clang-format -i -n --Werror` and run `lint-clang`? And commit new changes?

Already done, it should be ok now. @kirillzyusko

kirillzyusko (approved):

kirillzyusko:
@zhouzh1 hm, fails again... Did it pass locally?

kirillzyusko:
Thank you for your contribution @zhouzh1 ! ❤️

zhouzh1:
> @zhouzh1 hm, fails again... Did it pass locally?

Yes I did pass the `yarn lint-clang` check locally, not sure why it still failed on the CI..., anyway, thanks for your fix.

kirillzyusko:
My pleasure! 😊

PR #1061, opened 2025-08-01
docs: remove a note that library is incompatible with Expo Go

kirillzyusko:
### JS

-
-

### iOS

-
-

### Android

-
-

## 📝 Checklist

- [ ] CI successfully passed

### DIFF
--- a/docs/docs/installation.mdx
+++ b/docs/docs/installation.mdx
@@ -44,10 +44,6 @@ npm install react-native-keyboard-controller --save
 npx expo install react-native-keyboard-controller
 ```
 
-:::note Only Expo Dev client compatible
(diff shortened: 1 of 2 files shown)


jerone:
Came here to mention this. This package is now part of Expo Go. See https://github.com/expo/expo/pull/37942

kirillzyusko:
Thanks @jerone 

But when I try to run it in Expo Snack I get this:

image

Is it only a problem of Expo snack? Do you use Expo Go with this library and it works well for you?

jerone:
I tried it for a second in Expo Go and it worked just fine.
Because of planning and priorities I cannot test it fully.

kirillzyusko:
Thanks @jerone 

I found the problem why it doesn't work in Expo snacks and fixed it in https://github.com/kirillzyusko/react-native-keyboard-controller/pull/1209 🤞 

I'll merge this PR, starting from version `1.20.x` the lib will be fully compatible with expo 😎

Issue #1141, opened 2025-10-05
KeyboardAvoidingView does not work inside Modals if "prefer Cross-Fade Transitions" are enabled


buschco:
**Describe the bug**
The KeyboardAvoidingView is not expanding

**Code snippet**
```tsx
import React, { useState } from "react";
import {
  Modal,
  StyleSheet,
  Text,
  TextInput,
  TouchableOpacity,
  View,
} from "react-native";
import { ScrollView } from "react-native-gesture-handler";
import { KeyboardAvoidingView } from "react-native-keyboard-controller";
import { useSafeAreaInsets } from "react-native-safe-area-context";

import type { ViewProps } from "react-native-svg/lib/typescript/fabric/utils";

const styles = StyleSheet.create({
  flex1: { flex: 1 },
  flexGrow1: { flexGrow: 1 },
  heading: {
    color: "black",
    fontSize: 36,
    marginBottom: 48,
    fontWeight: "600",
  },
  inner: {
    padding: 24,
    flex: 1,
    justifyContent: "space-between",
  },
  textInput: {
    height: 45,
    borderColor: "#000000",
    borderWidth: 1,
    borderRadius: 10,
    marginBottom: 36,
    paddingLeft: 10,
  },
  button: {
    marginTop: 12,
    height: 45,
    borderRadius: 10,
    backgroundColor: "rgb(40, 64, 147)",
    justifyContent: "center",
    alignItems: "center",
  },
  text: {
    fontWeight: "500",
    fontSize: 16,
    color: "white",
  },

  backgroundColorRed: {
    backgroundColor: "red",
  },
});

function C() {
  return (
    <KeyboardAvoidingView
      behavior="padding"
      contentContainerStyle={styles.flex1 as ViewProps["style"]}
      keyboardVerticalOffset={100}
      style={styles.flex1}
      testID="keyboard_avoiding_view.container"
    >
      <ScrollView
        alwaysBounceVertical={false}
        automaticallyAdjustsScrollIndicatorInsets={false}
        contentContainerStyle={styles.flexGrow1}
        contentInset={{ bottom: 30 }}
        keyboardShouldPersistTaps="handled"
        scrollEventThrottle={300}
        scrollIndicatorInsets={{
          right: 1,
          bottom: 30,
        }}
        style={styles.flex1}
      >
        <View style={styles.inner}>
          <Text style={styles.heading}>Header</Text>
          <View>
            <TextInput
              placeholder="Username"
[...]

PR #1158, opened 2025-10-13
fix: iOS 26 + Modal

kirillzyusko:
## 📜 Description

Fixed a problem when events stop working on iOS 26 inside `Modal`.

## 💡 Motivation and Context

The current problem is that we attach a `keyboardLayoutGuide` view to the `rootView`. However when non-transparent `Modal` gets shown, then `rootView` will be detached and modal root view will be attached:

|App content|Modal shown|
|------------|-------------|
|image|image|

The first attempt that I tried to do is to attach `keyboardLayoutGuide` directly to the window. However in this case:
- I can not see `keyboardLayoutGuide` view
- it looks like it's not getting animated.

So I'm not sure if it's actually working method (maybe that `keyboardLayoutGuide` gets overwritten later).

Anyway I found the other way. When keyboard appear we'll receive a notification `UIWindow.didBecomeVisibleNotification` - from the handler of this notification we can attach a `keyboardLayoutGuide` and it will be animated when keyboard appears (even after initial attachment - if we would use `keyboardWillShow` then this trick wouldn't work, because it's too late to attach something there and rely on animated transitions - I think `keyboardLayoutGuide` internally also relies on keyboard notifications and if we attach it from keyboard notification then newly mounted handler will not receive an event).

To assure better performance I decided to keep a reference to the view to which handler has been attached. It allows us to skip re-creation in case if view hasn't been changed.

Initial changes for https://github.com/kirillzyusko/react-native-keyboard-controller/issues/1141

### iOS

- added subscription to `UIWindow.didBecomeVisibleNotification`;
- attach `keyboardLayoutGuide` tracking view from `didBecomeVisibleNotification` notification handler;
- added a mechanism that allows to avoid unnecessary view re-creations.

Tested manually on iPhone 16 Pro (iOS 26, XCode 26).

|Before|After|
|-------|----|
|<video src="https://github.com/user-attachments/assets/bc678ec2-ecd8-4e47-99fe-aa2dfb4a7e3d">|<video src="https://github.com/user-attachments/assets/7ca19b2b-ab0e-4e73-8f49-8381bf6b421b">|

## 📝 Checklist

- [x] CI successfully passed

### DIFF
--- a/ios/observers/movement/KeyboardTrackingView.swift
+++ b/ios/observers/movement/KeyboardTrackingView.swift
@@ -14,6 +14,7 @@ import UIKit
 final class KeyboardTrackingView: UIView {
   private var keyboardView: UIView? { KeyboardViewLocator.shared.resolve() }
   private var keyboardHeight = 0.0
+  private weak var currentAttachedView: UIView?
 
   static let invalidPosition: CGFloat = -.greatestFiniteMagnitude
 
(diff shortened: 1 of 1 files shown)


OwenMelbz:
Will this affect `KeyboardStickyView` ?

This was working fine until iOS 26 came around causing chaos - the apps not been touched in maybe a year or so.

image

Essentially got something like...

```
<View>
  <KeyboardGestureArea interpolator="ios">
    <ScrollView>
      <TextInput />
    </ScrollView>
    <KeyboardStickyView>
       <View bg="red" />
    </KeyboardStickyView>
  </KeyboardGestureArea>
</View>
```

kirillzyusko:
@OwenMelbz the `KeyboardStickyView` relies only on `onStart` event 🤔 And from what I saw `onStart` event gets dispatched properly (but `onEnd` gets messed up), so based on that info I can assume that `KeyboardStickyView` should work well?..

Can you prepare a repo showing the problem? (I also see that `interactive` event may drive values returned by `useKeyboardAnimation` hook, and I made some changes in `1.19.1`)

So please, test with latest `1.19.1` version and if the issue is reproducible, then open a new issue 🙏

Issue #1182, opened 2025-10-29
[iOS] [Expo] KeyboardAwareScrollView and KeyboardAvoidingView works in debug build, but not --variant Release


NicTorgersen:
**Describe the bug**
This library doesn't seem to work with iOS 26 and `--variant Release` (or a production build using `eas`), with Expo 54.

It works fine in iOS 18.5. I see various other issues in the issue tracker regarding iOS 26 inconsistencies, but for me it just doesn't work at all.

Works fine on Android in Release-mode.

It seems like only animating the scroll in the view is what breaks, because padding is being applied to the bottom of the view (I can scroll to the bottom with the entire page visible above the keyboard).

**Code sample**
Now, the app I am working on is quite the quilt. 

This is App.tsx:
```
<SafeAreaProvider>
    <KeyboardProvider>
        <GestureHandlerRootView style={{ flex: 1 }}>
            <ReduxProvider store={store}>
                <DependencyProvider>
                    <NativeBaseProvider isSSR={false}>
                        <NavigationContainer theme={AEKTheme}>
                            <AppNavigation />
                        </NavigationContainer>
                    </NativeBaseProvider>
                </DependencyProvider>
                <StatusBar style="dark" />
            </ReduxProvider>
        </GestureHandlerRootView>
    </KeyboardProvider>
</SafeAreaProvider>
```

The views are roughly like follows:
```
<>
    <LoadingIndicator isVisible={isLoading} />
    <KeyboardAwareScrollView 
        bottomOffset={Platform.OS === "ios" ? insets.bottom : 0} 
        ref={pageScrollView} 
        style={{ flex: 1, padding: 10 }}
    >
        ... tons of content
    </KeyboardAwareScrollView>
<>
```

**Repository With the Bug**
In this example the app is much more succint and simple:  
[https://github.com/nictorgersen/react-native-keyboard-controller-ios-26-issue](https://github.com/nictorgersen/react-native-keyboard-controller-ios-26-issue)

Had an AI set up the basics, but the issue is there on my iPhone with iOS 26 (only in Release-mode, never in Debug).

**To Reproduce**

1. Set up a new expo-project (expo 54 at the time of writing)
2. Install `react-native-keyboard-controller` (expo adds 1.18.5 by default, manually upgrading to 1.19.3 proves indifferent)
3. Set up a basic page with `KeyboardAwareScrollView` or `KeyboardAvoidingView`
[...]

PR #1184, opened 2025-10-31
fix: non-working release version on iOS 26

kirillzyusko:
## 📜 Description

Fixed an issue when `react-native-keyboard-controller` is not reporting keyboard movements when launching from release app version on iOS 26.

## 💡 Motivation and Context

The problem is that if we attach `keyboardLayoutGuide` to `rootView` straight in mount lifecycle, then `rootView` may be not available yet. To fix this problem we need to subscribe to event which will indicate that `rootView` is available and only then attach a view.
 
Closes https://github.com/kirillzyusko/react-native-keyboard-controller/issues/1182

### iOS

- attach `KeyboardTrackingView` only when app became active;

Tested manually on iPhone 16 Pro (iOS 26.1, real device).

|Before|After|
|-------|-----|
|<video src="https://github.com/user-attachments/assets/ff7c8982-8d19-4e9d-a09f-718f5b67c33b">|<video src="https://github.com/user-attachments/assets/9b2fc7ad-d702-4f6c-ba3d-a8932e7c5fa7">|

## 📝 Checklist

- [x] CI successfully passed

### DIFF
--- a/ios/observers/movement/KeyboardTrackingView.swift
+++ b/ios/observers/movement/KeyboardTrackingView.swift
@@ -56,7 +56,12 @@ final class KeyboardTrackingView: UIView {
       name: UIResponder.keyboardDidShowNotification,
       object: nil
     )
-    attachToTopmostView()
+    NotificationCenter.default.addObserver(
+      self,
+      selector: #selector(attachToTopmostView),
+      name: UIApplication.didBecomeActiveNotification,
+      object: nil
+    )
   }
 
   override func willMove(toWindow newWindow: UIWindow?) {


joacub:
now it does not work in development or TestFligh

it work after losing the focus of the app and coming back.

kirillzyusko:
@joacub eventually you are right 😐 I can reproduce the issue in example project. Could you please open a new issue so that I don't forget about this issue? 🙏 

P. S. I should enable e2e tests for iOS 26 - so many issues 😓

joacub:
> @joacub eventually you are right 😐 I can reproduce the issue in example project. Could you please open a new issue so that I don't forget about this issue? 🙏
> 
> P. S. I should enable e2e tests for iOS 26 - so many issues 😓

I already applied the changes. I don’t believe this is on your side — it seems more related to the numerous updates and changes Apple has recently made to their APIs.

Issue #1191, opened 2025-11-03
now it does not work in development or testflight


joacub:
now it does not work in development or TestFligh

it work after losing the focus of the app and coming back.

kirillzyusko:
For context: it was working before `1.19.3` on iOS 26 and stopped after `1.19.4`

joacub:
> For context: it was working before `1.19.3` on iOS 26 and stopped after `1.19.4`

Yes, it was working in version 1.19.3, but in the new version the behavior is much better. In 1.19.3, using KeyboardAwareScrollView caused noticeable jumping in some screens — not consistently, but enough to impact the experience. It seemed to improve slightly after interacting with the same screen multiple times, but the new version definitely handles it far more smoothly.

gabrielcalistro:
I don't know if its the same issue but there's something happening since I upgraded version. It was reported to me on 1.19.2, but it was only happening in production. Now, after 1.19.4, its happening in simulator as well. 

There is a noticeable delay before the UI updates to accommodate the keyboard.

I've started a new project to test it clean.
So, here some context: 

```
import { StatusBar } from "expo-status-bar";
import { StyleSheet, Text, TextInput, View } from "react-native";
import {
  KeyboardAvoidingView,
  KeyboardProvider,
} from "react-native-keyboard-controller";

export default function App() {
  return (
    <KeyboardProvider>
      <KeyboardAvoidingView style={{ flex: 1 }} behavior="padding">
        <View style={styles.container}>
          <Text>Open up App.js to start working on your app!</Text>
          <TextInput placeholder="Type anything here" />
          <StatusBar style="auto" />
        </View>
      </KeyboardAvoidingView>
    </KeyboardProvider>
  );
}

const styles = StyleSheet.create({
  container: {
    flex: 1,
    backgroundColor: "#fff",
    alignItems: "center",
    justifyContent: "center",
  },
});

```

System:
      OS: macOS 15.6.1
      Shell: 5.9
    Binaries:
      Node: 22.17.0
      npm: 10.9.2
      Watchman: 2025.10.20.00
    Managers:
      CocoaPods: 1.16.2
    SDKs:
      iOS SDK:
        Platforms: DriverKit 25.0, iOS 26.0, macOS 26.0, tvOS 26.0, visionOS 26.0, watchOS 26.0
    IDEs:
      Android Studio: 2024.3 AI-243.26053.27.2432.13536105
      Xcode: 26.0.1/17A400
[...]

PR #1193, opened 2025-11-04
fix: attaching `KeyboardTrackingView`

kirillzyusko:
## 📜 Description

Fixed a problem of non working `onMove` handler in development builds.

## 💡 Motivation and Context

In https://github.com/kirillzyusko/react-native-keyboard-controller/pull/1184 I started to rely that `didBecomeActive` event will be caught and we can attach the view from this listener. But turns out that in dev builds we miss this event. So at the moment:
- view hierarchy may be not ready by time of creating `KeyboardTrackingView` (noticeable in release versions);
- after `didBecomeActive` hierarchy is ready (but we may miss this event if we subscribed later, and also we rely on sequence of Apple events and it's kind of transitive dependencies that may break at any point of time).

So I decided to re-work the approach. We attach `KeyboardControllerView` anyway, so I thought it's good to add a side effect in `didMoveToWindow` and attach our `KeyboardTrackingView`. For that I added optional param to `attachToTopmostView` method (called `toWindow/window`). If we pass the window then we use that param, otherwise rely on `activeWindow` and take `rootViewController.view`.

Such combination seems to solve all these issues when `onMove` stops working under certain conditions. Now we always pass a window reference for a first mount and don't rely on internal Apple timing and events.

Closes https://github.com/kirillzyusko/react-native-keyboard-controller/issues/1191

### E2E

- sync assets with working `onMove` handler on iOS 26;

### iOS

- make `KeyboardTrackingView` public;
- pass `toWindow/window` into `attachToTopmostView` method;
- pass `window` into `attachToTopmostView` from `didMoveToWindow` lifecycle.

Tested:
- iPhone 16 Pro, iOS 26 (XCode 16, physical device) - repro app (release/debug + cold start); ✅
- iPhone 16 Pro, iOS 26 (XCode 16, simulator) - example app (release/debug + cold start) + Modal in separate window example; ✅ 
- iPhone 16 Pro, iOS 26 (XCode 26, simulator) - example app (release/debug + cold start) + Modal in separate window example; ✅

|Before (debug)|After (debug)|Release repro|
|-------|----|--------------|
|<video src="https://github.com/user-attachments/assets/9af9992c-462c-45fc-8d2e-51d2fbe66dd9">|<video src="https://github.com/user-attachments/assets/7dcabd42-7310-4251-9870-29ba7c073517">|<video src="https://github.com/user-attachments/assets/68bd986f-7c1e-414f-a492-9178d72d5b8a">|

## 📝 Checklist

- [x] CI successfully passed

### DIFF
--- a/ios/observers/movement/KeyboardTrackingView.swift
+++ b/ios/observers/movement/KeyboardTrackingView.swift
@@ -66,7 +66,7 @@ public final class KeyboardTrackingView: UIView {
   }
 
   @objc public func attachToTopmostView(toWindow window: UIWindow? = nil) {
-    guard let topView = (window ?? UIApplication.shared.activeWindow)?.rootViewController?.view else { return }
+    guard let topView = (window?.rootViewController ?? UIApplication.topViewController())?.view else { return }
 
     if currentAttachedView === topView { return }
 


e2e/kit/assets/ios/iPhone 17 Pro/AwareScrollViewSecondInputFocused.png
kirillzyusko: This is actually a bug spotted by e2e tests that was introduced in `1.19.4`. The `onMove` handler wasn't working, so `KeyboardAwareScrollView` wasn't functioning properly 😓

joacub:
waiting for the rlease to release our app

kirillzyusko:
@joacub the fix has been published in `1.19.5` version 🎉

PR #1260, opened 2025-12-29
Call correct onInterceptHoverEvent super in onInterceptHoverEvent()

MegaManSec:
## 📜 Description

Fix a copy/paste mistake in `OverKeyboardRootViewGroup.onInterceptHoverEvent`. The method incorrectly delegates to `super.onHoverEvent(event)` instead of `super.onInterceptHoverEvent(event)`, which bypasses the proper intercept path for hover events.

This change updates the return statement to call the correct superclass method.

## 💡 Motivation and Context

`onInterceptHoverEvent` is specifically intended to decide whether hover events should be intercepted before reaching child views. Calling `super.onHoverEvent` from the intercept override is incorrect and can lead to inconsistent hover/pointer behavior (including accessibility/hover dispatch) compared to the expected ViewGroup event flow.

This appears to be a straightforward copy/paste error given that `onHoverEvent` already calls `super.onHoverEvent(event)`.

`react-native` PR for reference: https://github.com/facebook/react-native/pull/54991

### Android

* Fixed `onInterceptHoverEvent` to call `super.onInterceptHoverEvent(event)` instead of `super.onHoverEvent(event)`.

### DIFF
--- a/android/src/main/java/com/reactnativekeyboardcontroller/views/overlay/OverKeyboardViewGroup.kt
+++ b/android/src/main/java/com/reactnativekeyboardcontroller/views/overlay/OverKeyboardViewGroup.kt
@@ -197,7 +197,7 @@ class OverKeyboardRootViewGroup(
     eventDispatcher?.let {
       jsPointerDispatcher?.handleMotionEventCompat(event, it, true)
     }
-    return super.onHoverEvent(event)
+    return super.onInterceptHoverEvent(event)
   }
 
   override fun onHoverEvent(event: MotionEvent): Boolean {


kirillzyusko:
Thank you @MegaManSec for your PR! Let's wait for green CI and I'll merge it straight away!

kirillzyusko:
Actually the same copy/paste error is present in `DialogRootViewGroup` (react-native codebase). You may want to submit a PR to RN core as well 😊 

Could you please confirm, that screen reader (with enabled accessibility) works fine with these changes?

MegaManSec:
I didn't actually test these changes: it was picked up during a source code review.

I will report this to facebook/react-native too, thank you.

kirillzyusko:
Okay, I'll test it then before merge 👍 

For e2e tests I'm using detox, and this framework doesn't use a11y tree for testing (unlike Maestro, for example), so I'll conduct additional testing 👍

Issue #1298, opened 2026-01-29
Invalid `onInteractive` event when pressing outside of `TextInput` to close keyboard


chrispader:
**Describe the bug**

I'm using `useKeyboardHandler` to react to the keyboard and translate scrollable content in a list. When i press outside of the focused `TextInput` without dragging the screen first an invalid `onInteractive` event is triggered with a too small keyboard height value (`e.height`).

**Code snippet**
Add your code snippet where error has been occurred.

```
const height = useSharedValue(0);

useKeyboardHandler({
  onInteractive: (e) => {
    'worklet';

    height.set(e.height);
  },
});
```

**Repo for reproducing**

https://github.com/Expensify/App/tree/36ce4dab66cc59b4c2e831f2b19376075495384f

**To Reproduce**

1. Open the keyboard by focussing/pressing a `TextInput` on the screen.
2. Wait for the keyboard to open
3. Close the keyboard by pressing outside of the `TextInput`

**Expected behavior**
There should be an `onStart` event first and no `onInteractive` (Also the `e.height` value in `onInteractive` should be exactly or approximately the max height of the keyboard, not way lower)

**Screenshots**

https://github.com/user-attachments/assets/f6464740-4a74-4882-8f50-f90c9a61ec33

**Smartphone (please complete the following information):**

- Desktop OS: macOS 26.2
- Device: iPhone 17 Pro Simulator
- OS: iOS 26.2 (Xcode 26.2)
- RN version: 0.81.4
- RN architecture: New Architecture/Fabric
- JS engine: Hermes
- Library version: 1.20.7

PR #1299, opened 2026-01-30
fix: wrong `onInteractive` event when keyboard closed on iOS 26+ with attached `KeyboardGestureArea`

kirillzyusko:
## 📜 Description

Fixed undesired `onInteractive` event during keyboard dismissal (non interactive).

## 💡 Motivation and Context

It looks like on iOS 26 timings are different than on previous iOS versions and now when we detach `InvisibleAccessoryView` we may get a KVO event listener (that will report invalid keyboard frame).

To avoid that I added a condition that will ignore this interactive event if you dismiss keyboard with attached `InvisibleAccessoryView`.

Closes https://github.com/kirillzyusko/react-native-keyboard-controller/issues/1298

### iOS

- use `KeyboardEventsIgnored` in KVO;

Tested manually in example project.

|Before|After|
|-------|-----|
|image|image|

## 📝 Checklist

- [x] CI successfully passed

### DIFF
--- a/ios/observers/movement/observer/KeyboardMovementObserver+Interactive.swift
+++ b/ios/observers/movement/observer/KeyboardMovementObserver+Interactive.swift
@@ -30,6 +30,10 @@ extension KeyboardMovementObserver {
       return
     }
 
+    if KeyboardEventsIgnorer.shared.shouldIgnore {
+      return
+    }
+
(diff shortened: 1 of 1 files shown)


chrispader (approved):
This change worked for me! Thanks

PR #1304, opened 2026-02-06
fix: non-working interactive keyboard dismissal

kirillzyusko:
## 📜 Description

Fixed an issue with non-working `onInteractive` handlers.

## 💡 Motivation and Context

The main issue has been introduced here: https://github.com/kirillzyusko/react-native-keyboard-controller/pull/1161

And the issue became visible very well after https://github.com/kirillzyusko/react-native-keyboard-controller/pull/1299

### 1️⃣ `shouldIgnore` was never becoming `false`

After

```swift
      activeTextInput.inputAccessoryView = currentInputAccessoryView
      activeTextInput.reloadInputViews()
```

We were receiving two events:
- `keyboardWillShow`
- `keyboardDidShow`

And in previous implementation we simply ignored `keyboardWillShow` and in `keyboardDidShow` were updating internal variables (resetting `shouldIgnore` to `false`, updating real keyboard frame, etc.) but we weren't dispatching the event to JS.

Now, when we don't receive `keyboardDidShow` and dispatch it manually from `keyboardWillShow` it breaks the logic. Now we can't simply ignore `keyboardWillShow` - instead we should replicate the logic of `keyboardDidShow`. This is the first fix.

### 2️⃣ Wrong `progress` value

We shouldn't simply move:

```swift
      guard !KeyboardEventsIgnorer.shared.shouldIgnore else {
        KeyboardEventsIgnorer.shared.shouldIgnoreKeyboardEvents = false
        return
      }
```

To the top, because we will not update real keyboard/keyboard frame, and because of that keyboard progress will be `1.21` etc. We need to update internal values and only then block event propagation.

### 3️⃣ Wrong `onEnd` event after keyboard appear after interactive dismissal

We had this line:

```swift
let height = self.keyboardHeight - KeyboardAreaExtender.shared.offset
```

It's incorrect, because we already do this operation in:

```swift
  var keyboardHeight: CGFloat {
    get { _keyboardHeight - KeyboardAreaExtender.shared.offset }
    set { _keyboardHeight = newValue }
  }
```

Originally this code was a ternary operator: 

```swift
      // if the event is caught in between it's highly likely that it could be a "resize" event
      // so we just read actual keyboard frame value in this case
      let height = timestamp >= didShowDeadline ? self.keyboardHeight : position - KeyboardAreaExtender.shared.offset
```

We did `-` operation only with `position` (that value included additional offset). So I removed double `-` - we should just read `self.keyboardHeight` after a re-writing operation and it should give us correct values. 

### 4️⃣ Missing `onEnd` event after keyboard disappear

It happened because we schedule/cancel `end` events in two places:
- duration setter;
- animation setter.

We always cancel previous animation, even though new setter may not set a new animation. And it's also incorrect behavior, because on iOS 26 when interactive keyboard gets shown (after interactive dismissal) the duration may be 0, but `animation` may be `nil` (iOS bug). In this case `duration` will schedule `end` event that later will be cancelled by `animation` (and it's a problem, because we'll never pause display link and it'll cancel all `onInteractive` events).

In this PR I decided to schedule calls in centralized way (from `will` listeners). Even if animation duration is not initialized we'll use default `duration` anyway, so it's safe now.

I tested cases with plain animation, with interactive dismissal, with switching keyboard to emojis etc. and haven't found any issues. I tested it on various iOS versions (iOS 26, iOS 17, iOS 16).

<hr />

Correct fix for https://github.com/kirillzyusko/react-native-keyboard-controller/issues/1298

> Note: this may require preparation of `1.20.8` because interactive keyboard handling is broken in `1.20.7`

### iOS

- move code with resetting `shouldIgnore` from `keyboardDidAppear` into `keyboardWillAppear`;
- don't apply double `minus` operation when deriving keyboard height in `keyboardDidAppear` handler;

Tested manually on:
- iPhone 17 Pro (iOS 26.2)
- iPhone 16 Pro (iOS 18.5)
- iPhone 15 Pro (iOS 17.5)
- iPhone 14 Pro (iOS 16.4)

|iOS 17|iOS 26|
|------|-------|
|<video src="https://github.com/user-attachments/assets/b1cbe98c-a3e2-41a2-a7b6-9d4dac2bab92">|<video src="https://github.com/user-attachments/assets/6dc7d5d8-7b78-4c60-ad03-b6155e3aa3a0">|

## 📝 Checklist

- [x] CI successfully passed

### DIFF
--- a/ios/observers/movement/observer/KeyboardMovementObserver+Interactive.swift
+++ b/ios/observers/movement/observer/KeyboardMovementObserver+Interactive.swift
@@ -48,7 +48,7 @@ extension KeyboardMovementObserver {
     }
 
     prevKeyboardPosition = position
-    
+
     keyboardDidTask?.cancel()
 
     onEvent(


vinceprofeta:
When can we expect this to be released?

kirillzyusko:
Hey @vinceprofeta 

It's published under 1.21.0-beta.1 tag - feel free to try and let me know if you encounter any issues ☺️

NepeinAV:
@kirillzyusko hi, I'm facing a problem with a FlatList wrapped in a KeyboardAvoidingView. when the keyboard is open and I swipe on FlatList, the bottom indentation freezes for a while. before this PR, the indentation did not disappear at all.

Repo: https://github.com/NepeinAV/ChatIsolation

https://github.com/user-attachments/assets/c8d366b3-55ee-48bb-85f1-86947eebb917

kirillzyusko:
Hey @NepeinAV 

Could you please open a new issue and attach all information there? 🙏

I am afraid that I may just forget about it (fixing too many problems in parallel and just can't hold everything in my head)

PR #1308, opened 2026-02-10
fix: call `maybeScroll` in `onEnd` handler in `KeyboardAwareScrollView`

kirillzyusko:
## 📜 Description

Call `maybeScroll` in `onEnd` handler.

## 💡 Motivation and Context

The `onMove` can be missing because of many things:
- resize event;
- enabled a11y with reduced motion on Android;
- iOS 15 may not have enough `onMove` keyboard events on first app start.

So in this PR I added `maybeScroll` call in `onEnd`. Before we were avoiding this, because on iOS we could have a situation when `end` event arrives before `start` but it has been fixed in https://github.com/kirillzyusko/react-native-keyboard-controller/pull/1161 so it should be safe to do now 🤞 

### JS

- call `maybeScroll(e, false)` in `onEnd` handler'

Tested via e2e tests 🤞 

|Before|After|
|-------|-----|
|image|image|

## 📝 Checklist

- [x] CI successfully passed

### DIFF
--- a/src/components/KeyboardAwareScrollView/index.tsx
+++ b/src/components/KeyboardAwareScrollView/index.tsx
@@ -462,6 +462,8 @@ const KeyboardAwareScrollView = forwardRef<
         onEnd: (e) => {
           "worklet";
 
+          maybeScroll(e.height);
+
           removeGhostPadding(e.height);
 
           keyboardHeight.value = e.height;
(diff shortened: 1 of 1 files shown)


kirillzyusko:
It doesn't fix the flakiness of iOS 15. Moreover it introduces regressions on Android 9, so I'm going to close this PR.

PR #1338, opened 2026-03-04
fix(ios): guard delayed manual keyboard did events by current keyboard state

litinskii:
## 📜 Description

Fix race condition in manual iOS `did` keyboard events introduced in #1161.

When navigating between screens, keyboard events may arrive in non-linear order (`willHide` -> `didShow` -> `didHide`). The last delayed `didHide` could incorrectly win and reset keyboard height to `0`, even when keyboard is already visible on the new screen.

This change keeps the manual `did` strategy from #1161, but validates actual keyboard state at delayed execution time.

## 💡 Motivation and Context

After upgrading to `1.20.6`, some navigation flows started producing incorrect keyboard offset on the next screen:

- keyboard is visible
- but computed offset is `0`
- content goes under keyboard

Root cause: delayed `DispatchWorkItem` tasks may execute out of logical order during responder/navigation transitions.

This PR prevents stale delayed events from overriding current keyboard state.

### Problem Statement (clear event order issue)

When navigating between screens, keyboard events can arrive in an invalid order:

1. `keyboard hide`
2. `keyboard show end`
3. `keyboard hide end`

As a result, the last processed event is `hide`, keyboard height is set to `0`, and offset is recalculated to `0`.
On the next screen, keyboard is visible, but bottom inset is not applied, so content goes under the keyboard.

### Reproduction

```ts
Keyboard.dismiss();

await wait(400);

// navigate to screen with auto-focus input
navigation.navigate(...);
```

Expected:
- final keyboard state should match visible keyboard (`height > 0`)
- bottom offset should remain applied on the destination screen

Actual before fix:
- delayed stale `didHide` may run last
- keyboard height/offset becomes `0` while keyboard is still visible

### iOS

- validate delayed `did` event against current keyboard visibility before emitting `didShow`/`didHide`
- add stale-task guard (`keyboardDidEventID`) so old delayed tasks cannot override newer state
- cancel pending delayed `did` tasks and clean up watchers/KVO on `unmount`
- keep manual `did` dispatch approach from #1161 (no rollback to system `keyboardDid*`)

Tested manually on iOS navigation scenario with focused input:

1. Open screen A with focused input (keyboard visible).
2. Navigate quickly to screen B where input becomes focused.
3. Reproduce event interleaving where hide/show callbacks race.
4. Verify final state:
   - keyboard remains visible
   - `keyboardDidHide` does not incorrectly win
   - keyboard height/offset is not reset to `0`
   - content stays above keyboard on the new screen

Also verified `unmount` path does not emit late delayed `did` events from removed screen observer.

Before

https://github.com/user-attachments/assets/a69c43b3-11ce-4e98-a79f-fffe279b3bc3

After

https://github.com/user-attachments/assets/6df95ff8-355f-46c8-abc8-ef428963b6d2

## 📝 Checklist

- [ ] CI successfully passed

### DIFF
--- a/ios/observers/movement/observer/KeyboardMovementObserver+Lifecycle.swift
+++ b/ios/observers/movement/observer/KeyboardMovementObserver+Lifecycle.swift
@@ -29,6 +29,11 @@ public extension KeyboardMovementObserver {
 
   @objc func unmount() {
     isMounted = false
+    keyboardDidTask?.cancel()
+    keyboardDidTask = nil
+    notification = nil
+    removeKeyboardWatcher()
+    removeKVObserver()
     // swiftlint:disable:next notification_center_detachment
     NotificationCenter.default.removeObserver(self)
   }
--- a/ios/observers/movement/observer/KeyboardMovementObserver+Listeners.swift
+++ b/ios/observers/movement/observer/KeyboardMovementObserver+Listeners.swift
@@ -6,6 +6,32 @@
 //
 
 extension KeyboardMovementObserver {
+  private func emitKeyboardDidAppear(height: CGFloat, duration: Int) {
+    let progress = 1.0
+
+    onCancelAnimation()
+    onEvent("onKeyboardMoveEnd", height as NSNumber, progress as NSNumber, duration as NSNumber, tag)
+    onNotify("KeyboardController::keyboardDidShow", buildEventParams(height, duration, tag))
+
+    removeKeyboardWatcher()
+    setupKVObserver()
+    animation = nil
+  }
+
+  private func emitKeyboardDidDisappear(duration: Int) {
+    onCancelAnimation()
+    onEvent("onKeyboardMoveEnd", 0 as NSNumber, 0, duration as NSNumber, tag)
+    onNotify("KeyboardController::keyboardDidHide", buildEventParams(0, duration, tag))
+
+    removeKeyboardWatcher()
+    animation = nil
+  }
+
+  private func currentKeyboardHeight() -> CGFloat {
+    let (visibleKeyboardHeight, _) = keyboardTrackingView.view.frameTransitionInWindow
+    return max(CGFloat(visibleKeyboardHeight) - KeyboardAreaExtender.shared.offset, 0)
+  }
+
   @objc func keyboardWillAppear(_ notification: Notification) {
     guard !UIResponder.isKeyboardPreloading else { return }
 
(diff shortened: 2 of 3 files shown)


kirillzyusko:
Hey @litinskii 

Is this problem still actual even for `1.21.0-beta.3`? Shouldn't `didShow` event simply cancel previous `willHide` schedule?

litinskii:
1.20.7 with bug

https://github.com/user-attachments/assets/a69c43b3-11ce-4e98-a79f-fffe279b3bc3

1.20.5

https://github.com/user-attachments/assets/407abc60-2d2c-4a81-807f-6ddff7dae347

1.20.7 with this PR

https://github.com/user-attachments/assets/6df95ff8-355f-46c8-abc8-ef428963b6d2

1.21.0-beta.3 is it bug or it just start working in another way 

https://github.com/user-attachments/assets/2b90b03a-4c39-4402-a5a5-783b673ad244

Thanks, that makes sense. I tested 1.21.0-beta.3 from npm (not from a local git tag/branch), so I can’t point to an exact repo SHA from my local clone.

This can still happen in 1.21.0-beta.3 (in another way i suppose) in navigation/auto-focus flows, because didShow can only cancel willHide if both schedules belong to the same active observer instance and lifecycle window.

In this race, the delayed hide is often scheduled by the previous screen observer, then the next screen mounts and schedules show. Those tasks can complete out of order, and a stale didHide may still run last after transition, resetting height/offset to 0 while keyboard is visible.

So “didShow cancels previous willHide” is necessary, but not sufficient for this case.
In this PR I try to validates actual keyboard visibility at execution time and cancels pending tasks on unmount.

PR is focused on the stale delayed-event race during navigation/auto-focus:
`hide -> showEnd -> hideEnd`, where a stale `didHide` can still win and reset height/offset to `0` while keyboard is visible.

So “didShow cancels previous willHide” is helpful, but not always sufficient across observer lifecycle boundaries (previous screen vs next screen).

Could you share the exact commit SHA used for 1.21.0-beta.3? I don’t see that tag/branch in my local clone.

This is the smallest fix I found for this specific race, but I agree there may be a better long-term design.
I’m happy to rework the PR if you suggest a cleaner approach.

kirillzyusko:
> Could you share the exact commit SHA used for 1.21.0-beta.3?

Sure, this is commit: https://github.com/kirillzyusko/react-native-keyboard-controller/commit/11c0dbfba864a1e5fdff51e673b7e2dd407fc6ea

> I don’t see that tag/branch in my local clone.

Yes, because I forgot to create them 😬 

Thank you for videos! It helps a lot! I'll try to analyze problem in more details (I had only a quick look). Thank you again for raising this issue and preparing solution ❤️

kirillzyusko:
BTW @litinskii  Don't you mind to switch to [discord](https://discordapp.com/users/973642097888489512) (nickname is kiryl.ziusko) for further communication?

I think communication in messenger will be much faster and more productive 👀

kirillzyusko:
Discussed it in Discord and confirmed that the issue is not reproducible with latest version of the library, so I'm closing this PR.

Thank you for you contribution @litinskii 🙌

PR #1339, opened 2026-03-04
chore: release 1.21.0

kirillzyusko:
# Release

Version 1.21.0

### DIFF
--- a/FabricExample/ios/Podfile.lock
+++ b/FabricExample/ios/Podfile.lock
@@ -1836,7 +1836,7 @@ PODS:
     - ReactCommon/turbomodule/core
     - SocketRocket
     - Yoga
-  - react-native-keyboard-controller (1.20.7):
+  - react-native-keyboard-controller (1.21.0):
     - boost
     - DoubleConversion
     - fast_float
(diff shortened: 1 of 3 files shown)


d4rky-pl:
@kirillzyusko I know I'm being rude right now and I'm really sorry about this but I'm in the middle of migrating to this library and current master likely has the fix I need - any idea if the release ETA is in days or weeks/months? 😅

kirillzyusko:
@d4rky-pl I think you may try to use `1.21.0-beta.3` release, it's almost identical to current `main` branch, I believe 🤞

kirillzyusko:
@d4rky-pl but answering on your original question - I expect to publish a new version next week 👀

d4rky-pl:
@kirillzyusko amazing, thanks! Turns out the bug I was hoping a fix for was an issue with React Native not your library after all 🙃 so thankfully it's not a blocker for our release anymore.

Still looking forward, your library is quite cool and definitely beats `react-native-keyboard-aware-scroll-view` we've been using for the past few years in terms of UX

branaust:
just wanted to say this release is fixing so many issues i was experiencing. contantly say frame rates dropping by like 20 frames on the UI. this fixes it completely. also keyboard handling and focus are much better. thank you for your work

kirillzyusko:
Thank you @branaust ❤️ 

I'm planning to release it on Monday next week (just need to finish blogpost part). Other than that the release is ready to enter "stable" phase 😊

PR #1346, opened 2026-03-08
fix: add `automaticOffset` prop for correct KAV positioning in modals

### LINKED ISSUE #867: KeyboardStickyView not moving with keyboard within modal
Thanks again for your hard work on this amazing project!

**Describe the bug**
When using the KeyboardStickyView or KeyboardToolbar it doesnt move with the keyboard on scrolling the modal up/down

**Code snippet**
```
const SettingModalWrapper = ({
  label,
  isModalVisible,
  setIsModalVisible,
  children,
}: SettingModalWrapperProps) => {
  const form = useForm({
    onSubmit: async ({ value }) => {
      console.log(value)
    },
    defaultValues: {
      email: "",
    },
  })

  return (
    <Modal
      animationType="slide"
      presentationStyle="pageSheet"
      visible={isModalVisible}
      onRequestClose={() => setIsModalVisible(false)}
    >
      <View className="h-full bg-primary-200">
        <KeyboardAvoidingView className="px-5 py-5 flex-1">
          <form.Field
            name="email"
            children={(field) => (
              <View className="">
                <TextInput
                  value={field.state.value}
                  onBlur={field.handleBlur}
                  onChangeText={field.handleChange}
                  placeholder="example@toaster.com"
                  spellCheck={true}
                  textContentType="emailAddress"
                />
              </View>
            )}
          />
        </KeyboardAvoidingView>
        <KeyboardToolbar />
      </View>
    </Modal>
  )
}
```

**Repo for reproducing**
https://github.com/spaansba/Toaster

**To Reproduce**
1. Add a modal to your app
2. Add a textinput
3. Click the textinput and move the modal

**Expected behavior**
The toolbar should move with the keyboard when moving the modal up/down

**Screenshots**

https://github.com/user-attachments/assets/732e80c2-c00d-4ec5-8d9b-12602a3a5695

- Device: iPhone 13
- OS: IOS 18.3.1
- RN version: 0.76.7
- expo version: 52.0.35
- RN architecture: new
- Library version: 1.16.7

thomasttvo:
## Problem

`KeyboardAvoidingView` doesn't work correctly in `pageSheet` modals **on iOS** because `onLayout` returns parent-relative coordinates. In a modal, the parent's `y=0` doesn't correspond to screen `y=0`, so the keyboard overlap calculation uses incorrect positions — resulting in the keyboard covering the input or the view being offset incorrectly.

This is an iOS-specific issue. On Android, modals use a separate `Dialog` window, but `measureInWindow` (backed by `getLocationOnScreen()`) already returns correct absolute screen coordinates regardless of which window the view is in.

## Solution

Add an opt-in `automaticOffset` prop (default `false`) that gates the `measureInWindow` behavior. This preserves backward compatibility — existing users who set `keyboardVerticalOffset` to their header height continue to work unchanged.

When `automaticOffset={true}`:
- Uses `measureInWindow` instead of `onLayout` to get absolute screen coordinates, so the view's position is correctly detected in modals, behind navigation headers, etc.
- `keyboardVerticalOffset` becomes purely **additive** extra space rather than compensation for unknown positioning — the default of `0` works correctly out of the box.

When `automaticOffset={false}` (default):
- Behavior is identical to before this PR — uses `onLayout` parent-relative coordinates, and `keyboardVerticalOffset` must be set to the header height manually.

Uses the existing `useCombinedRef` hook to maintain both the internal ref (needed for `measureInWindow`) and the forwarded ref from the consumer.

## Example app

Added a new **"KeyboardAvoidingView Automatic"** example screen (`KeyboardAvoidingViewAutomatic`) to showcase `automaticOffset` behavior:
- Demonstrates all three behavior modes (padding, height, position)
- Auto/Manual toggle to compare `automaticOffset={true}` vs `automaticOffset={false}`
- Includes a `pageSheet` Modal to test the modal positioning fix
- Configurable `keyboardVerticalOffset` toggle (+0, +50, +100)
- Reusable `KAVContent` component shared between regular screen and modal

No changes to the existing KAV example screen or E2E tests.

## Test Plan

**iOS (iPhone 16 simulator), `automaticOffset={true}`, `keyboardVerticalOffset={0}`:**
- ✅ Regular screen — all content visible above keyboard (padding, position modes)
- ✅ Modal — all content visible above keyboard (padding, position modes)

**`automaticOffset={false}` (default), `keyboardVerticalOffset={100}`:**
- ✅ Regular screen — same behavior as before this PR
- ✅ RN implementation — same behavior as before this PR

Fixes #867

### DIFF
--- a/src/components/KeyboardAvoidingView/index.tsx
+++ b/src/components/KeyboardAvoidingView/index.tsx
@@ -137,9 +137,6 @@ const KeyboardAvoidingView = forwardRef<
           initialFrame.value === null ||
           behavior !== "height"
         ) {
-          // When the keyboard is open in height mode, storing the shrunk
-          // frame would cause a feedback loop (resize → recalculate →
-          // resize), so we skip that case entirely.
           // eslint-disable-next-line react-compiler/react-compiler
           initialFrame.value = layout;
         }


src/components/KeyboardAvoidingView/index.tsx
```diff
@@ -177,6 +196,20 @@ const KeyboardAvoidingView = forwardRef<
           return {};
       }
     }, [behavior, enabled, interpolateToRelativeKeyboardHeight]);
+    const combinedRef = useCallback(
```
kirillzyusko: Better to use `useCombinedRef` hook that we have in this package 👀

kirillzyusko:
Thank you @thomasttvo for your PR

Unfortunately I can not merge it as is. Try to open example app and check the difference between package/RN implementations. You will see that package implementation will push the content significantly higher than it's expected to do. Ideally RN and package implementation should produce identical visual result.

This is a breaking change and we should understand what's causing this and fix it before a merge 🤞 

You can also download e2e artifacts to see what exactly doesn't work...

thomasttvo:
hey @kirillzyusko sorry about that! I meant to put this PR back in a WIP state since there's some significant changes to it since it's opened. Let me check what's going on with these issues and get back to you. Thank you for your patience!

thomasttvo:
@kirillzyusko ready for review!

I tested everything manually and verified it works locally, but I may have missed a few details. Let me know if you spot anything or have any feedback on the changes.

kirillzyusko:
Hey @thomasttvo 

I reviewed your PR and I think it has breaking changes. Since we start to calculate `keyboardVerticalOffset` partially on our own -> all users that update from previous version will have breaking changes (i. e. the content will be pushed higher than expected).

I tend to think this is a breaking change and I don't want to created a release V2 just because of these changes. Do you think we can hide new functionality under feature flag/new prop?

For example `automaticallyCalculateVerticalKeyboard`/`includeAbsoluteScreenPosition`/whatever you-or-claude think is better and when it's true, then use new path for calculation?

Again, my concern is that `onLayout` and `measureInWindow` may produce different results. `onLayout` may not include `Header`/`SafeArea` etc. height. -- `measureInWindow` will account that spacing. Maybe we can add a new internal property like `absoluteY` which will be `0` by default, and when new prop is passed then we'll update it (via measure function) and we will always include it in calculations?

thomasttvo:
makes sense @kirillzyusko let me see what I can do

thomasttvo:
@kirillzyusko I decided to go with `automaticOffset`. I also had to change the example screen a bit, because there are too many buttons on the `headerRight` now, and I added some new tests. Let me know what you think. 

Some of the automated tests were failing. I haven't had a chance to dig into them, but I'm guessing it was due to the fact that it was a breaking change. There's also the argos [failure](https://app.argos-ci.com/kirillzyusko/react-native-keyboard-controller/builds/631/271136408), which probably wasn't caused by this PR

kirillzyusko:
> There's also the argos [failure](https://app.argos-ci.com/kirillzyusko/react-native-keyboard-controller/builds/631/271136408), which probably wasn't caused by this PR

You can ignore them for now. Argos job is flacky a little bit and needs to be fixed (I'll handle it separately)

thomasttvo:
@kirillzyusko 🤦 , this is very strange. I can see the `y=0` on initial load but now it quickly changes to a positive value the moment the modal finishes the animation, which means we don't need the extra logic. It wasn't doing that before, maybe it was just me hallucinating, or things got weird after rounds of hot reloads. I removed all the code that workarounds the `y=0` for `height` and `position` mode, and added a separate screen for the Automatic Offset mode. Let me know what you think!

kirillzyusko (approved):
Looks good! Thank you for your contribution ❤️

kirillzyusko:
FYI @thomasttvo it seems like it doesn't work with new architecture (FabricExample) + Modal:

image

Paper architecture works well. Is it a bug in `measureInWindow`?..

thomasttvo:
@kirillzyusko on it! should have checked that we've been testing against Paper all this time.

kirillzyusko:
@thomasttvo the problem seems to be in react-native and how `measureInWindow` is implemented on a new architecture. I'll prepare a fix in RN repo, no worries!

kirillzyusko:
I opened a PR here: https://github.com/facebook/react-native/pull/56062

PR #1371, opened 2026-03-16
Fix extraContentPadding not adjusting scroll when it changes by large amount

trcoffman:
## 📜 Description

If your chat composer changes height by a small amount, such as adding an additional line, `extraContentPadding` works fine. However, if you paste a large blob of text in, the scroll position adjustment doesn't happen and your content gets occluded by the chat composer.

## 💡 Motivation and Context

It makes `extraContentPadding` work the way that it should.

## 📢 Changelog

KeyboardChatScrollView: Fix `extraContentPadding` not adjusting scroll consistently.

### JS

- 
-

### iOS

- Use contentOffset in implementation of extraContentPadding to synchronously adjust scroll position
-

### Android

- Use requestAnimationFrame to delay scrollTo until next frame after extraContentPadding change gets committed

Manually tested on an iPhone (iPhone 16 Pro) and Android device (Pixel 7)

### Before

iOS:

https://github.com/user-attachments/assets/0404e3c5-9dc3-4d55-b3b7-9483b1db2677

Android:

https://github.com/user-attachments/assets/b31a2966-033f-4f04-9196-31ecf44d0e17

### After

iOS:

https://github.com/user-attachments/assets/188ced9f-08f5-4952-a525-b996ed3673ac

Android:

https://github.com/user-attachments/assets/e9438991-0414-4743-948d-6e81a6ce044b

## 📝 Checklist

- [ ] CI successfully passed

### DIFF
--- a/src/components/KeyboardChatScrollView/useExtraContentPadding/__fixtures__/setup.ts
+++ b/src/components/KeyboardChatScrollView/useExtraContentPadding/__fixtures__/setup.ts
@@ -12,6 +12,14 @@ export let reactionEffect: (current: number, previous: number | null) => void;
 
 export const flushRAF = () => new Promise((resolve) => setTimeout(resolve, 0));
 
+let mockForceLegacy = false;
+
+jest.mock("../../../../architecture", () => ({
+  get IS_FABRIC() {
+    return !mockForceLegacy;
+  },
+}));
+
 jest.mock("react-native-reanimated", () => ({
   ...require("react-native-reanimated/mock"),
   scrollTo: (...args: unknown[]) => mockScrollTo(...args),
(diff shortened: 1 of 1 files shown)


kirillzyusko (changes requested):

kirillzyusko:
I left only one comment! Other than that the PR looks solid and I think we can merge it!

kirillzyusko:
Right now on Android there is a significant delay when input grows vs when scroll happens. I think it's caused by many factors: async events to JS, additional frame wait etc. I know you @trcoffman have some ideas how to improve that, so looking forward to your second PR!

kirillzyusko:
@trcoffman sorry, on iOS I found one bug:

https://github.com/user-attachments/assets/79d8e8c5-e0c0-475a-9882-bc86a69a43ea

Android doesn't have this bug

trcoffman:
> @trcoffman sorry, on iOS I found one bug:
> 
>  Simulator.Screen.Recording.-.iPhone.17.Pro.-.2026-03-19.at.10.08.45.mov 
> Android doesn't have this bug


https://github.com/user-attachments/assets/10cc1c4b-49f5-451d-b0c4-717b730b1928

I wasn't able to reproduce this on the AI chat screen.

The chat composer on the Chat Kit screen never expands for me. Something is really wrong with this chat composer I think.

Issue #1375, opened 2026-03-17
KeyboardAwareScrollView causes 'Could not find view for tag' errors in v1.21.0


eyedroot:
## Description

After upgrading from `react-native-keyboard-controller@1.20.7` to `1.21.0`, the `KeyboardAwareScrollView` component causes multiple `Could not find view for tag` promise rejection errors on screen mount.

## Environment

- **react-native**: 0.83.2
- **react-native-keyboard-controller**: 1.21.0 (works fine on 1.20.7)
- **expo**: 54
- **react-native-pager-view**: 8.0.0
- **react-native-tab-view**: 4.3.0
- **New Architecture (Fabric)**: enabled
- **Platform**: iOS 26.3 (Simulator, iPhone 17 Pro)

## Steps to Reproduce

1. Use `KeyboardAwareScrollView` inside `react-native-tab-view`'s `TabView` scenes
2. Each tab scene wraps its form content with `KeyboardAwareScrollView`
3. Navigate to the screen containing the TabView

## Expected Behavior

No errors on screen mount.

## Actual Behavior

Multiple `Could not find view for tag` promise rejection errors appear immediately on mount. The number of errors corresponds to the number of `KeyboardAwareScrollView` instances and/or TextInput fields rendered across the tab scenes.

```
ERROR [Error: Uncaught (in promise, id: 0) Error: Could not find view for tag]
ERROR [Error: Uncaught (in promise, id: 1) Error: Could not find view for tag]
ERROR [Error: Uncaught (in promise, id: 2) Error: Could not find view for tag]
ERROR [Error: Uncaught (in promise, id: 3) Error: Could not find view for tag]
ERROR [Error: Uncaught (in promise, id: 4) Error: Could not find view for tag]
ERROR [Error: Uncaught (in promise, id: 5) Error: Could not find view for tag]
ERROR [Error: Uncaught (in promise, id: 6) Error: Could not find view for tag]
```

## Verification

- Replacing `KeyboardAwareScrollView` with a plain `ScrollView` eliminates the errors completely.
- Downgrading to `react-native-keyboard-controller@1.20.7` also resolves the issue.
- The errors occur regardless of `TabView`'s `lazy` prop (tested with and without).

## Minimal Example

```tsx
// Each tab scene uses KeyboardAwareScrollView
<KeyboardAwareScrollView
  style={{ flex: 1 }}
  contentContainerStyle={{ padding: 16, paddingBottom: 128 }}
[...]

PR #1379, opened 2026-03-18
fix: could not fing view for tag warning

kirillzyusko:
## 📜 Description

Silent `could not fing view for tag` warning for `KeyboardAwareScrollView` (similarly how it works for `KeyboardAvoidingView` now).

## 💡 Motivation and Context

This issue got introduced after https://github.com/kirillzyusko/react-native-keyboard-controller/pull/1352 (which was based on https://github.com/kirillzyusko/react-native-keyboard-controller/pull/1346)

In certain cases `onLayout` may be triggered but view may be not laid out yet (if you use it with `react-native-pager` lazy). Since changes in https://github.com/kirillzyusko/react-native-keyboard-controller/pull/1352 were purely additional (we try to understand the position of element on the screen, but before these changes we always assumed we were in the top of the screen) I decided to wrap it for now in `try/catch` block.

Silenting error is not a correct fix, but this is the best what we can do now - we will not spam logs when devs update the package (in https://github.com/kirillzyusko/react-native-keyboard-controller/pull/1346 we also used `catch` block), but yes, in certain cases detection of top border will not work (and it didn't work before, so it's fair enough to silent the error).

Closes https://github.com/kirillzyusko/react-native-keyboard-controller/issues/1375

### JS

- handle rejected promises within `try/catch` block;

Tested via e2e tests, I didn't test changes manually because I don't have a repro.

## 📝 Checklist

- [x] CI successfully passed

### DIFF
--- a/src/components/KeyboardAwareScrollView/index.tsx
+++ b/src/components/KeyboardAwareScrollView/index.tsx
@@ -164,11 +164,15 @@ const KeyboardAwareScrollView = forwardRef<
         onLayout?.(e);
 
         if (handle !== null) {
-          const { y } = await KeyboardControllerNative.viewPositionInWindow(
-            handle,
-          );
+          try {
+            const { y } = await KeyboardControllerNative.viewPositionInWindow(
+              handle,
+            );
 
-          scrollViewPageY.value = y;
+            scrollViewPageY.value = y;
+          } catch {
+            // ignore
+          }
         }
       },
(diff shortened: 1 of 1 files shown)


Junveloper:
This was tearing my hair out! After seeing your tweet, I've updated to 1.21 to implement chat and whoa I thought I did something!! So glad I checked the PR! Thank you as always.

Issue #1318, opened 2026-02-20
Prevent vertically centered TextInput from jumping when keyboard resizes


calypsow777:
The following is more of a question than a feature request.

Say I have a vertically centered `TextInput` inside a `KeyboardAwareScrollView`:

```jsx
<KeyboardAwareScrollView
        contentContainerStyle={{ flex: 1, padding: 20, justifyContent: 'center' }}>

        <TextInput
            style={{ fontSize: 24 }}
            autoComplete="email"
            placeholder="Your e-mail address"/>
</KeyboardAwareScrollView>
```

This might happen for example in an onboarding flow where screens have a single input that we would like to highlight.

Some devices display and hide an extra bar at the top of the keyboard, depending on the input context, e.g. to show autocomplete suggestions.
This leads to the keyboard's height increasing and decreasing, and this in turn leads to the `TextInput` "jumping" to recenter itself in the available space.

Here's a video example :

https://github.com/user-attachments/assets/c2c7007b-ddf8-4cc5-b0a3-65fc69b65f5c

Now this behavior is expected, and I don't have issues with it personally.
The problem is that some end users don't really notice that extra bar appearing and disappearing; all they notice is a `TextInput` jumping around, and they conclude that the UI is buggy.

To curb the problem, I was thinking of implementing this rather fussy behavior:

1. Center the `TextInput` when the keyboard appears the first time, as it does now.
2. Ignore subsequent keyboard resizes, unless the resize causes the keyboard to (partially) hide the `TextInput`. Ignoring resizes that don't hide the `TextInput` means that it won't always be perfectly centered, but that's OK.

Is there an easy way to achieve this with this library? I checked the docs for `KeyboardAvoidingView`/`KeyboardAwareScrollView`/`KeyboardStickyView` but I don't see anything that would help.

kirillzyusko:
Hey @calypsow777 

Could you check if the issue is fixed on `1.21.0-beta.1`/`1.21.0-beta.2`

kirillzyusko:
Oh, re-read it more carefully. I don't think ignoring resize is a good idea, because user may switch to emoji keyboard and it has very big height 🤔

calypsow777:
Hey; thanks for your quick feedback.

> I don't think ignoring resize is a good idea, because user may switch to emoji keyboard and it has very big height

Indeed, that's why ideally it would only ignore resizes that don't cover the `TextInput`.

[...]

PR #1407, opened 2026-03-27
fix: wrong offset for `KeyboardAwareScrollView` when keyboard resizes

kirillzyusko:
## 📜 Description

Fixed an issue when resizing keyboard constantly pushes content up.

## 💡 Motivation and Context

When a user switches between keyboard types (e.g. text keyboard → emoji keyboard and back), `KeyboardAwareScrollView` was accumulating scroll offset instead of converging to a stable position. Each toggle would push the scroll further, even though the input's actual on-screen position hadn't changed. It's caused by 2 bugs:

### 1️⃣ Stale `absoluteY` during keyboard resize

`maybeScroll` uses `layout.value.absoluteY` to determine how much to scroll. This value is populated by the native `FocusedInputLayoutChanged` event. On iOS, the native observer uses `KVO` on the input's center relative to its superview (the scroll view's content area). When the scroll view scrolls, the input's center relative to the content doesn't change — only the scroll view's bounds.origin shifts — so `KVO` never fires and absoluteY is never updated. 

As a result, when the emoji keyboard opens and the component scrolls (say) `+76px`, the stored `absoluteY` still reflects the pre-scroll position. When switching back to the text keyboard, `maybeScroll` computes an overshoot based on that stale value, producing a cumulative drift on every toggle.

On Android the native side does report updated `absoluteY` on scroll, so only iOS is affected — but the stale-value case is latent everywhere.

### 2️⃣ `scrollBeforeKeyboardMovement` was clobbered on every keyboard resize

This value is used as the interpolation base when the keyboard hides: we scroll back to it. It was being overwritten on every `onStart` event, including keyboard-size-only changes (emoji ↔ text). That meant after one toggle, the "home position" shifted, and hiding the keyboard would scroll to the wrong place.

<hr />

The fix consist from two parts.

For 1️⃣ problem - when `onStart` fires for a keyboard-size change on the same input (`keyboardWillChangeSize && !actualFocusChanged`), we recompute absoluteY from first principles:

```ts
absoluteY = layout.y - position.value + scrollViewPageY
```

`layout.y` is the input's Y offset relative to the scroll view's content — this is stable and never stale. Subtracting the current scroll offset and adding the scroll view's on-screen origin gives the true current position on screen, regardless of what the native side reported.

For 2️⃣ problem - we gate the `scrollBeforeKeyboardMovement` update behind `actualFocusChanged`. Keyboard-size-only changes (same target, different height) no longer disturb this value, so the scroll-back-on-hide always returns to the position captured at the original focus event.

Closes https://github.com/kirillzyusko/react-native-keyboard-controller/issues/1318

### JS

- added `actualFocusChanged` variable inside `onStart` handler;
- update `scrollBeforeKeyboardMovement` only if `actualFocusChanged===true` (not `focusWasChanged`);
- covered new code with unit tests;
- update `layout.value.absoluteY` when keyboard gets resized;

Tested paper arch on:
- Pixel 9 Pro (API 35)
- iPhone 17 Pro (iOS 26.2)

|iOS|Android|
|-------|-----|
|<video src="https://github.com/user-attachments/assets/7c61ccca-5257-4fe7-b751-d0e22574f757">|<video src="https://github.com/user-attachments/assets/6c8f98dc-e471-4050-b27a-0cdc840b6a87">|

## 📝 Checklist

- [x] CI successfully passed

### DIFF
--- a/src/components/KeyboardAwareScrollView/index.tsx
+++ b/src/components/KeyboardAwareScrollView/index.tsx
@@ -432,21 +432,15 @@ const KeyboardAwareScrollView = forwardRef<
             }
 
             // save current scroll position - when keyboard will hide we'll reuse
-            // this value to achieve smooth hide effect.
-            // Only for actual focus changes — keyboard-size-only changes (e.g.
-            // emoji ↔ text toggle) should preserve the original value so that
-            // absoluteY correction and keyboard-hide scroll-back work correctly.
+            // this value to achieve smooth hide effect (only for actual focus change
+            // keyboard resize handled below)
             if (actualFocusChanged) {
               scrollBeforeKeyboardMovement.value = position.value;
             }
           }
 
-          // When keyboard size changes for the same input (e.g. emoji ↔ text),
-          // absoluteY in the stored layout may be stale on some platforms (iOS
-          // doesn't update it when scroll changes, Android does). Compute the
-          // actual on-screen position from the input's content-relative Y,
-          // current scroll offset, and ScrollView's position — this is correct
-          // regardless of whether the native side updated absoluteY.
+          // compute the actual on-screen position from the input's content-relative Y,
+          // current scroll offset, and ScrollView's position (case for keyboard resize)
           if (keyboardWillChangeSize && !actualFocusChanged && layout.value) {
             layout.value = {
               ...layout.value,


kirillzyusko:
Closing in favour of https://github.com/kirillzyusko/react-native-keyboard-controller/pull/1417

PR #1408, opened 2026-03-27
docs: fix KeyboardAwareScrollView Reanimated example

litinskii:
## Summary

This updates the `KeyboardAwareScrollView` + `@gorhom/bottom-sheet` docs example to rely on Reanimated type inference instead of passing `KeyboardAwareScrollViewProps` explicitly.

## Why

With newer `react-native-reanimated` typings, this form:

```ts
Reanimated.createAnimatedComponent<KeyboardAwareScrollViewProps>(
  KeyboardAwareScrollView,
);
```

can resolve against the wrong overload and produce a `FlatList`-related TypeScript error, for example:

```ts
Argument of type 'ForwardRefExoticComponent<...>' is not assignable to parameter of type 'typeof FlatList<...>'
```

This does not seem to be a runtime issue with `react-native-keyboard-controller` itself. It is a typing mismatch caused by newer Reanimated overload resolution.

## Fix

The example now uses:

```ts
Reanimated.createAnimatedComponent(KeyboardAwareScrollView);
```

This lets Reanimated infer the component props automatically and keeps the example compatible across Reanimated versions.

## Notes

I only updated the current docs example and added a short explanation near the snippet so it is clearer why the explicit generic is avoided.

### DIFF
--- a/docs/docs/api/components/keyboard-aware-scroll-view.mdx
+++ b/docs/docs/api/components/keyboard-aware-scroll-view.mdx
@@ -212,8 +212,9 @@ import {
 import type { BottomSheetScrollViewProps } from "@gorhom/bottom-sheet/src/components/bottomSheetScrollable/types";
 import Reanimated from "react-native-reanimated";
 
-const AnimatedScrollView =
-  Reanimated.createAnimatedComponent(KeyboardAwareScrollView);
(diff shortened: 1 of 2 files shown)


kirillzyusko:
@litinskii could you please fix prettier? `yarn format` should do the work 🤞

Issue #1430, opened 2026-04-10
Expose `isAtEnd` from `KeyboardChatScrollView` for jump-to-latest and follow-mode UIs


jkinggg:
**Is your feature request related to a problem? Please describe.**

`KeyboardChatScrollView` already computes whether the scroll view is effectively at the end for `keyboardLiftBehavior="whenAtEnd"`, but that state is currently internal.

For chat-style UIs, apps often need the same signal for:

- showing or hiding a "scroll to bottom" / "jump to latest" button
- enabling follow-mode while streaming new assistant content
- tracking whether unread content is accumulating below the viewport

Today the app has to re-implement that logic outside the component. That is awkward because `KeyboardChatScrollView` already owns the relevant scroll metrics and keyboard-aware padding behavior, so app-side heuristics can drift from the component's internal definition of "at end".

**Describe the solution you'd like**

Expose an optional `isAtEnd?: SharedValue<boolean>` prop on `KeyboardChatScrollView`.

When provided, the component would update that shared value on the UI thread using the same internal scroll metrics and `isScrollAtEnd(...)` logic it already uses for `keyboardLiftBehavior="whenAtEnd"`.

Example:

```tsx
const isAtEnd = useSharedValue(true);

<KeyboardChatScrollView
  isAtEnd={isAtEnd}
  keyboardLiftBehavior="whenAtEnd"
/>
```

This keeps the public API narrow while still exposing the high-level state chat apps actually need.

**Describe alternatives you've considered**

1. Expose raw scroll position instead.

I think that is a less ideal API for this use case. Most consumers do not need raw `scrollY`; they need the semantic "is the user pinned to the latest content?" state. Exposing raw scroll metrics also leaks more implementation detail and pushes more edge-case handling into app code.

2. Recompute "at end" in the app.

This works imperfectly, especially once keyboard lift behavior, bottom padding, composer growth, and inverted/non-inverted behavior are involved. The component already has the authoritative inputs, so surfacing that result is more reliable.

3. Expose a callback instead of a shared value.

A callback is usable, but a `SharedValue<boolean>` fits the existing Reanimated-driven API surface better and avoids bouncing this state back to JS for consumers that want to animate directly from it.

**Additional context**

- This would be an additive API change.
- The implementation can be JS-only in `KeyboardChatScrollView`.
- Internally, the component already has everything needed: `scroll`, `layout`, `size`, and `isScrollAtEnd(...)`.
[...]

PR #1431, opened 2026-04-10
feat: expose `isAtEnd` shared value from `KeyboardChatScrollView`

jkinggg:
## 📜 Description

This PR adds an optional `isAtEnd?: SharedValue<boolean>` prop to `KeyboardChatScrollView`.

When provided, the component updates that shared value on the UI thread using the same internal scroll metrics and `isScrollAtEnd(...)` logic already used by `keyboardLiftBehavior="whenAtEnd"`.

The goal is to let chat-style apps build jump-to-latest / scroll-to-bottom affordances without reimplementing the component's internal "at end" logic in app code.

## 💡 Motivation and Context

This change is intended to solve a common gap in chat UIs:

- the component already knows whether the user is effectively at the end
- apps often need that same signal for follow-mode and jump-to-latest UI
- today, consumers must guess based on external scroll heuristics

That duplication is brittle because app-side heuristics can diverge from the component's own keyboard-aware behavior.

This PR keeps the API small by exposing the semantic state (`isAtEnd`) instead of lower-level raw scroll metrics.

Related issue: https://github.com/kirillzyusko/react-native-keyboard-controller/issues/1430

### JS

- add optional `isAtEnd?: SharedValue<boolean>` prop to `KeyboardChatScrollView`
- update `isAtEnd` on the UI thread using existing `isScrollAtEnd(...)` logic
- add TypeScript declarations for the new prop

### iOS

- no native changes

### Android

- no native changes

Validated in the fork clone with:

- `yarn typescript`
- `yarn lint src/components/KeyboardChatScrollView/index.tsx src/components/KeyboardChatScrollView/types.ts`

Not yet completed before opening this draft PR:

- `yarn test`
- manual verification in the example app on iOS
- manual verification in the example app on Android

No screenshots yet. This draft PR is mainly to support API discussion from #1430.

## 📝 Checklist

- [ ] CI successfully passed

### DIFF
--- a/src/components/KeyboardChatScrollView/types.ts
+++ b/src/components/KeyboardChatScrollView/types.ts
@@ -87,4 +87,11 @@ export type KeyboardChatScrollViewProps = {
    * Default is `undefined` (equivalent to `0` — no minimum floor).
    */
   blankSpace?: SharedValue<number>;
+  /**
+   * Optional shared value that tracks whether the scroll view is currently at the end.
+   *
+   * When provided, `KeyboardChatScrollView` updates it on the UI thread using the same
+   * scroll metrics that drive `keyboardLiftBehavior="whenAtEnd"`.
+   */
+  isAtEnd?: SharedValue<boolean>;
 } & ScrollViewProps;
--- a/src/components/KeyboardChatScrollView/index.tsx
+++ b/src/components/KeyboardChatScrollView/index.tsx
@@ -2,6 +2,7 @@ import React, { forwardRef, useCallback, useMemo } from "react";
 import { StyleSheet } from "react-native";
 import {
   makeMutable,
+  useAnimatedReaction,
   useAnimatedRef,
   useAnimatedStyle,
   useDerivedValue,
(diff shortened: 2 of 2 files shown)


kirillzyusko:
Hey @jkinggg 

Is it final PR? Can we move it to "ready for review" stage?

jkinggg (commented):
Hey @kirillzyusko this is ready

kirillzyusko:
Hey @jkinggg 

Thank you for PR! (And sorry for a long review, last week I was busy with other problems and couldn't allocate a time for supporting/developing this library).

Overall I'm not agains of having these changes in the lib. However one thing that concerns me is that we violate React rules by passing a mutable property... I think it's not really react style? Do you think we can pass custom callback, like `onAtEnd` (or any other name, I don't want to use `onEndReached` because this prop already exist in `FlatList`) and run this callback in reaction effect?

In this case it will be more customizable behavior, because developer may want to write custom functionality when end is reached (not only changing shared variable)?

What do you think about callback approach? 👀

src/components/KeyboardChatScrollView/types.ts
```diff
@@ -87,4 +87,11 @@ export type KeyboardChatScrollViewProps = {
    * Default is `undefined` (equivalent to `0` — no minimum floor).
    */
   blankSpace?: SharedValue<number>;
+  /**
+   * Optional shared value that tracks whether the scroll view is currently at the end.
+   *
+   * When provided, `KeyboardChatScrollView` updates it on the UI thread using the same
+   * scroll metrics that drive `keyboardLiftBehavior="whenAtEnd"`.
+   */
+  isAtEnd?: SharedValue<boolean>;
```
kirillzyusko, line 96: We also will need to update documentation (next version and current 1.21), but I can do it for you if you prefer 😊

kirillzyusko:
Hey @jkinggg 

So what do you think about the `callback` approach? 👀

PR #1455, opened 2026-05-11
KeyboardChatScrollView: Fix scroll gestures on Android when scrollview content is shorter than the scrollview's viewport

### LINKED ISSUE #1453: KeyboardChatScrollView (blankSpace): On Android, scroll gesture doesn't work on content when is less tall than viewport.
**Describe the bug**
If you have a KeyboardChatScrollView, and the content inside that is shorter than the viewport, and you add enough blankSpace to make it scrollable, then the scroll gesture does not work when you place your finger on the content. It only works when you place your finger on the blank space below the content and scroll. This only happens if the content is short enough that the scrollview would not be scrollable without `blankSpace` being applied. If the content becomes long enough that the scrollview would be scrollable whether or not you have blankSpace then everything works flawlessly.

**Repo for reproducing**
https://github.com/trcoffman/react-native-keyboard-controller/tree/blankSpace-android-bug-repro

**To Reproduce**
- Check out my branch
- Run FabricExample
- Open 'Blank Space Android Gesture Bug' example.

**Expected behavior**
- Placing finger on messages at the top and scrolling scrolls the scrollview.

**Screenshots**

https://github.com/user-attachments/assets/f50241e8-6c3a-408b-ab89-67d5f4a04d44

- Pixel 7 (Android 16).

trcoffman:
## 📜 Description

Fixes #1453 

Reworks the Android blank-space / `extraContentPadding` / keyboard-lift
behavior in `KeyboardChatScrollView` so it matches iOS. Four separate bugs
were stacked on top of each other; fixing one at a time required fixing
the rest.

## 💡 Motivation and Context

The original symptom was that on Android, a drag gesture started inside
the message area was ignored when the conversation was shorter than the
viewport — only drags starting inside the blank-space region scrolled.
Investigating that uncovered a chain of related bugs in the Android
blank-space path: mid-stream snap-to-0 as AI replies grew, wrong
`isScrollAtEnd` / `visibleFraction` / max-scroll math, and wildly wrong
keyboard-lift math on short content. The fixes bring Android's behavior
in line with the iOS per-frame model, where `contentInset.bottom` cleanly
separates the applied inset from the natural content size.

### JS

- `useChatKeyboard`: introduced a `naturalContentHeight()` helper on
  Android to subtract the applied inset (`max(blankSpace, padding +
  extraContentPadding)`) from `size.value.height`. Android reports
  `contentSize` as `contentView.getHeight()` which now includes the
  decorator's extension, whereas iOS reports natural content size
  independent of `contentInset`. Fed the adjusted value into
  `isScrollAtEnd`, `clampScrollIfNeeded`, `getVisibleMinimumPaddingFraction`,
  `clampedScrollTarget`, and the `persistent`-close `maxScroll`.
- `useChatKeyboard`: replaced the `visibleFraction >= 1` binary gate in
  `onStart` with proportional absorption matching iOS:
  `absorbed = max(0, visibleFraction * blankSpace - extraContentPadding)`.
  Fixes short content being pushed off the top when the keyboard opened
  (blankSpace extends past the viewport on short content, so the
  fraction is never 1 and the gate was always skipping absorption).
- `useChatKeyboard`: replaced `minimumPaddingFractionOnOpen` with a
  `visiblePaddingOnOpen` snapshot so `onMove` uses the same iOS formula
  as `onStart` (`max(0, visiblePadding - extraContent)` instead of
  `(blankSpace - extraContent) * fraction`). The two formulas only
  agreed at `fraction === 1`; at partial fractions the composer clipped
  the last message's timestamp.
- `useChatKeyboard`: non-inverted `onMove` now skips `e.duration === -1`
  snap-back events (matching the inverted path), so dragging the scroll
  view with the keyboard open no longer snaps the scroll position back
  on finger-up.
- `useChatKeyboard`: `persistent`-close now uses
  `offsetBefore + actualOpenShift` instead of
  `offsetBefore + padding.value`, so closing the keyboard doesn't jump
  content up when blankSpace absorbed part of the keyboard height.
- `useEndVisible`: added an `appliedInset` option and a Platform-gated
  natural-content-height adjustment so `isScrollAtEnd` reports correctly
  on Android.
- `KeyboardChatScrollView`: moved the `totalPadding` derived value above
  `useEndVisible` and threaded it through as `appliedInset`.
- `ScrollViewWithBottomPadding`: minor comment cleanup clarifying that
  `contentInset` / `scrollIndicatorInsets` are iOS-only and
  `contentInsetBottom` / `contentInsetTop` are the Android
  `KeyboardControllerScrollView` props.

### Android

- `ClippingScrollViewDecoratorView`: the scroll-range extension now
  extends `contentView.bottom` directly (by the applied inset) instead
  of using `scrollView.paddingBottom`. Base Android
  `ScrollView.computeVerticalScrollRange()` derives from
  `child.getBottom()` and ignores padding, so `paddingBottom` alone
  didn't make `canScrollVertically(1)` return true inside the message
  area on short content — drags started there weren't intercepted.
  Extending `child.bottom` fixes that.
- `ClippingScrollViewDecoratorView`: installs an
  `OnLayoutChangeListener` on the content view that re-applies the
  `child.bottom` extension on every Yoga relayout, and re-orders RN's
  own listener to fire after ours (RN's listener clamps `scrollY`
  against the current `child.getHeight()` at `ReactScrollView.java:1275`
  — if it ran first it would use the un-extended natural height and
  snap `scrollY` toward 0 as each AI reply chunk arrived). The listener
  is cleaned up in `onDetachedFromWindow`.

### iOS

- No iOS native changes. JS-side math was adjusted to match iOS's
  existing per-frame model (see "JS" above).

## 📝 Checklist

- [ ] CI successfully passed

### DIFF
--- a/FabricExample/src/screens/Examples/AILegendListChat/index.tsx
+++ b/FabricExample/src/screens/Examples/AILegendListChat/index.tsx
@@ -30,29 +30,8 @@ type Message = {
 
 const createId = () => String(Date.now());
 
-const INITIAL_AI_TEXT = `Tip: Type 'a' for a short reply, 'b' for medium, 'c' for long, or 'd' for extra long. Any other text picks a random length.
-
-React Native virtualization is a performance optimization technique that's crucial for handling large lists efficiently. Here's how it works:
-
-1. **Rendering Only Visible Items**: Instead of rendering all items in a list at once, virtualization only renders the items that are currently visible on screen, plus a small buffer of items just outside the visible area.
-
-2. **Dynamic Item Creation/Destruction**: As you scroll, items that move out of view are removed from the DOM/native view hierarchy, and new items that come into view are created. This keeps memory usage constant regardless of list size.
-
-3. **View Recycling**: Advanced virtualization systems reuse view components rather than creating new ones, which reduces garbage collection and improves performance.
-
-4. **Estimated vs Actual Sizing**: The system uses estimated item sizes to calculate scroll positions and total content size, then adjusts as actual sizes are measured.
-
-5. **Legend List Implementation**: Legend List enhances this by providing better handling of dynamic item sizes, bidirectional scrolling, and maintains scroll position more accurately than FlatList.
-
-The key benefits are:
-- Constant memory usage regardless of data size
-- Smooth scrolling performance
-- Better handling of dynamic content
-- Reduced time to interactive
-
-This makes it possible to scroll through thousands of items without performance degradation, which is essential for modern mobile apps dealing with large datasets like social media feeds, chat histories, or product catalogs.
-
-Tip: Type 'a' for a short reply, 'b' for medium, 'c' for long, or 'd' for extra long. Any other text picks a random length.`;
+const INITIAL_AI_TEXT =
+  "Tip: Type 'a' for a short reply, 'b' for medium, 'c' for long, or 'd' for extra long. Any other text picks a random length.";
 
 const INITIAL_MESSAGES: Message[] = [
   {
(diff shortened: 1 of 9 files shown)


kirillzyusko:
I took a different approach in https://github.com/kirillzyusko/react-native-keyboard-controller/pull/1487

This PR caused a regression on Android (KeyboardToolbar screen), the new PR doesn't introduce a regression 

The fix will be published in `1.21.12`

PR #1458, opened 2026-05-12
security: avoid compromised release being publish through the poisoned cache

kirillzyusko:
## 📜 Description

Don't use `cache` for publish job.

## 💡 Motivation and Context

Based on https://github.com/actions/setup-node/issues/1445#issuecomment-4040130833

### CI

- do not use cache for publishing (to avoid poisoned cache vector attack);

There is no way to test it at the moment 😅 

## 📝 Checklist

- [x] CI successfully passed

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -21,5 +21,6 @@ jobs:
           # It increases the chance of a compromised release being publish
           # See https://github.com/actions/setup-node/issues/1445#issuecomment-4040130833
           # cache: yarn
+          package-manager-cache: false
       - run: yarn install --frozen-lockfile
       - run: npm publish --provenance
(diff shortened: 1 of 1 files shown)


.github/workflows/publish.yml
```diff
@@ -17,6 +17,9 @@ jobs:
         with:
           node-version: "24.x"
           registry-url: "https://registry.npmjs.org"
-          cache: yarn
```
slorber, line 20: Hey

I'm not 100% sure because the action doc is not super clear but I think it's better to explicitly add `package-manager-cache: false`

kirillzyusko: Thank you @slorber 

I think it makes sense to add it anyway! Thank you again for discovering this problem and suggesting more defensive techniques, appreciate it! 😊

PR #1460, opened 2026-05-12
fix: guard around IllegalState exception

kirillzyusko:
### JS

-
-

### iOS

-
-

### Android

-
-

## 📝 Checklist

- [ ] CI successfully passed


kirillzyusko:
Closing this PR - it prevents a crash but after that IME is never shown (which indicates a broken app state, so closing it). We shouldn't try to ignore the exception - we should try to understand it and fix it properly.

Issue #1454, opened 2026-05-09
On Mac Catalyst, using `KeyboardAwareScrollView` causes abnormal crashes.


dppo:
In React Native v0.85.3, when using `KeyboardAwareScrollView` together with `TextInput`, both tapping the `TextInput` and tapping the close button in `native-stack` will cause the app to crash unexpectedly. The same code works normally in React Native v0.83.

Since the default React Native v0.85.3 setup cannot run directly on Mac Catalyst, I used the workaround mentioned in [this issue](https://github.com/facebook/react-native/issues/55540) to make the project run on Mac Catalyst.

```
pushd ios/Pods/React-Core-prebuilt/React.xcframework/ios-arm64_x86_64-maccatalyst/React.framework
rm -r React Resources
ln -s Versions/Current/Resources Resources
ln -s Versions/Current/React React
pushd Versions && rm -r Current && ln -s A Current && popd
popd

pushd ios/Pods/ReactNativeDependencies/framework/packages/react-native/ReactNativeDependencies.xcframework/ios-arm64_x86_64-maccatalyst/ReactNativeDependencies.framework
rm -r ReactNativeDependencies Resources
ln -s Versions/Current/Resources Resources
ln -s Versions/Current/ReactNativeDependencies ReactNativeDependencies
pushd Versions && rm -r Current && ln -s A Current && popd
mv ReactNativeDependencies_*.bundle Resources
popd
```

<img width="1832" height="1038" alt="Image" src="https://github.com/user-attachments/assets/2d302b5d-9b6d-448f-99d9-a07f41a256cf" />

<img width="1770" height="984" alt="Image" src="https://github.com/user-attachments/assets/286d81b9-18cd-4d1d-ad3f-28f7c197c0c4" />

kirillzyusko:
Hey @dppo 

It looks like it's infinite recursion 🤔 

Is it only Mac Catalyst? A noob question: how can I enable Mac Catalyst in example project? (I would love to fix this problem)

kirillzyusko:
The fix is published under `1.21.8` version 🤞

dppo:
Thank you very much. After testing, version 1.21.8 has successfully resolved this issue.

PR #1469, opened 2026-05-21
fix(ios): avoid swift_dynamicCast crash walking subviews on Mac Catalyst

skrtdev:
## Summary

On Mac Catalyst running in the Mac idiom, `view.subviews` can contain host containers (e.g. `UIRemoteView` wrappers around Mac-side AppKit views) whose pointers cannot be bridged to `UIView` at the Swift level. Iterating them with the standard:

```swift
for subview in subviews { … }
```

trips a `swift_dynamicCast` SIGSEGV — which fires from this library's two subview-walking entry points whenever something on screen calls `-becomeFirstResponder` or asks for the input-field hierarchy.

Two fixes, both gated to `#if targetEnvironment(macCatalyst)` + a `userInterfaceIdiom == .mac` runtime check, so iOS / iPadOS / Catalyst-as-iPad take the existing code paths byte-for-byte.

### 1. `UIView.findFirstResponder()`

Route through UIKit's responder chain via `sendAction(_:to: nil, from: nil, for:)` instead of recursing through subviews. The nil-target send walks the responder chain and stops at the first responder, which records itself into a static slot via a small selector helper. No subview tree walk.

### 2. `ViewHierarchyNavigator` (`getAllInputFields`, `findTextInputInHierarchy`)

Add a private `safeSubviews(of:)` helper that, on Catalyst-in-Mac-idiom, enumerates subviews via Objective-C KVC (`view.value(forKey: \"subviews\") as? NSArray`) and casts each entry defensively to `UIView`, silently skipping non-`UIView` host pointers. Outside Catalyst it's a one-line `return view.subviews` — no behaviour or perf change.

The three subview-iteration sites (`findTextInputs` body, `isGroupRoot` branch, `findTextInputInHierarchy` direction loop) all switch to `safeSubviews(of:)`.

## Test plan

- [ ] Build for iOS Simulator — no diff in any iOS code path, no behaviour change.
- [ ] Build for Mac Catalyst with `userInterfaceIdiom == .mac` — both fixes compile cleanly.
- [ ] Tap into a `TextInput` on a Catalyst Mac build — no `swift_dynamicCast` crash; keyboard toolbar previous/next still navigates between fields.
- [ ] `findFirstResponder()` returns the focused field on Catalyst Mac.

### DIFF
--- a/ios/extensions/UIView.swift
+++ b/ios/extensions/UIView.swift
@@ -39,6 +39,27 @@ public extension UIView {
     if isFirstResponder {
       return self
     }
+    #if targetEnvironment(macCatalyst)
+    // On Mac Catalyst running in the Mac idiom, `view.subviews` can contain
+    // host containers (e.g. UIRemoteView wrappers around Mac-side AppKit
+    // views) whose pointers cannot be bridged to UIView at the Swift level.
+    // Iterating them with the standard `for subview in subviews` loop trips
+    // a `swift_dynamicCast` SIGSEGV whenever findFirstResponder() is called
+    // during normal UIKit operations such as -becomeFirstResponder.
+    // Route through UIKit's responder chain via sendAction(_:to: nil, …)
+    // instead, which locates the first responder without touching the
+    // subview tree.
+    if UIDevice.current.userInterfaceIdiom == .mac {
+      UIResponder._kbcCapturedFirstResponder = nil
+      UIApplication.shared.sendAction(
+        #selector(UIResponder._kbcRecordFirstResponder(_:)),
+        to: nil,
+        from: nil,
+        for: nil
+      )
+      return UIResponder._kbcCapturedFirstResponder as? UIView
+    }
+    #endif
     for subview in subviews {
       if let responder = subview.findFirstResponder() {
         return responder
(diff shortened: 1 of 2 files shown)


kirillzyusko:
Hey @skrtdev 

Aren't you trying to fix the problem that has been fixed in https://github.com/kirillzyusko/react-native-keyboard-controller/issues/1454? Which library version do you use when crash is reproducible?

skrtdev:
I'll check if my crash is still reproducible in 1.21.8

kirillzyusko:
Hey @skrtdev 

Did you have a chance to test it?

skrtdev:
Hey @kirillzyusko 

I just checked and it still crashes

kirillzyusko:
@skrtdev would it be possible to provide a repro with this crash? I tried to repro it in example app and there it works perfectly fine 🤷‍♂️

skrtdev:
I just checked again. It works. Idk what was happening before

kirillzyusko:
Okay, maybe a cache issue 🙈 

I'll close the PR then, if you don't mind ❤️ Feel free to report any other new issues that you discover 👀

skrtdev:
Thank you

PR #1477, opened 2026-06-01
fix(android): use composition for JSPointerDispatcherCompat

cortinico:
## 📜 Description

JSPointerDispatcher became a final Kotlin class in React Native 0.87+
(after the Java-to-Kotlin migration in https://github.com/facebook/react-native/pull/56910), so
JSPointerDispatcherCompat can no longer extend it.

While I'm restoring it in:
- https://github.com/facebook/react-native/pull/57022

Ideally we would want to reduce the API surface of RN apis. 

So here I'm switching from inheritance to composition by holding an internal
JSPointerDispatcher delegate and forwarding calls to it.

Looking for guidance on how to test this.

N/A

## 📝 Checklist

- [ ] CI successfully passed

### DIFF
--- a/android/src/main/java/com/reactnativekeyboardcontroller/views/overlay/JSPointerDispatcherCompat.kt
+++ b/android/src/main/java/com/reactnativekeyboardcontroller/views/overlay/JSPointerDispatcherCompat.kt
@@ -1,17 +1,23 @@
 package com.reactnativekeyboardcontroller.views.overlay
 
 import android.view.MotionEvent
+import android.view.View
 import android.view.ViewGroup
 import com.facebook.react.uimanager.JSPointerDispatcher
 import com.facebook.react.uimanager.events.EventDispatcher
 import java.lang.reflect.Method
 
 /**
- * Compat layer for `JSPointerDispatcher` interface for RN < 0.72
+ * Compat layer for `JSPointerDispatcher` interface for RN < 0.72.
+ *
+ * Uses composition instead of inheritance because `JSPointerDispatcher`
+ * became a final Kotlin class in RN 0.87+.
  */
 class JSPointerDispatcherCompat(
   viewGroup: ViewGroup,
-) : JSPointerDispatcher(viewGroup) {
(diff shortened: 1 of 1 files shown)


android/src/main/java/com/reactnativekeyboardcontroller/views/overlay/JSPointerDispatcherCompat.kt
```diff
@@ -1,17 +1,23 @@
...
 import android.view.MotionEvent
+import android.view.View
 import android.view.ViewGroup
 import com.facebook.react.uimanager.JSPointerDispatcher
 import com.facebook.react.uimanager.events.EventDispatcher
 import java.lang.reflect.Method
 
 /**
- * Compat layer for `JSPointerDispatcher` interface for RN < 0.72
+ * Compat layer for `JSPointerDispatcher` interface for RN < 0.72.
```
cortinico, line 11: @kirillzyusko also this is quote of an old version of RN. Perhaps the compat layer can go altogether?

kirillzyusko (approved):

android/src/main/java/com/reactnativekeyboardcontroller/views/overlay/JSPointerDispatcherCompat.kt
kirillzyusko: Yes @cortinico 

I will do it this year. I'm going to drop old architecture support and bump minimal RN version to RN 0.77 or something like this 🤞 

I know that RN has window support for only 3 versions, but I believe I don't have so much conditional code to support more RN versions, so I'm fine to leave with few additional `if`-statements 🙂

kirillzyusko:
Thank you @cortinico for your PR! I'll prepare next patch release right now 👍

kirillzyusko:
@cortinico new `1.21.9` version has been published on npm!

cortinico:
Thank you very much @kirillzyusko 🙏

PR #1493, opened 2026-06-09
fix: reconcile frozen chat keyboard padding

jmeistrich:
## Problem

In LegendList's keyboard integration when a user sends a message we want to do at the same time:
- Scroll the new message to the top
- Close the keyboard

So I am using `freeze` to disable KeyboardChatScrollView's scrolling to allow our scrollToEnd to run instead. But then that makes it never report an inset less than the keyboard height, because freezing made it stop calculating keyboard height. 

`KeyboardChatScrollView` currently returns early from keyboard handlers when `freeze` is true. If the keyboard closes while frozen, the hook can miss the terminal keyboard height and keep stale keyboard padding after `freeze` becomes false.

In that state, `onContentInsetChange` can stop at the previous keyboard padding instead of continuing down to `0`, even though `freeze` is already false.

## Screenshot

This is what it looks like before the fix, it reports inset shrinking while the blankSize shrinks, but to a minimum of keyboard size (288 in this case). So then there's a big bottomInset.

Simulator Screenshot - iPhone 17 Pro - 2026-06-09 at 16 49 28

## Solution

Keep tracking the latest keyboard height while frozen, but continue to skip layout and scroll writes during the frozen interval. When `freeze` transitions from true to false, recompute the effective padding from the latest observed keyboard height.

This keeps `freeze` scoped to applying keyboard-driven layout updates, not observing keyboard state.

## Notes

This intentionally reconciles padding only. It does not replay skipped scroll operations after unfreezing, because that could introduce visible jumps and is not needed for the stale inset issue.

## Test plan

- Verified manually in a Legend List chat repro on iOS and Android.
- `yarn test src/components/KeyboardChatScrollView/useChatKeyboard/__tests__/index.ios.spec.ts src/components/KeyboardChatScrollView/useChatKeyboard/__tests__/freeze.android.spec.ts --runInBand`
- `yarn lint --quiet src/components/KeyboardChatScrollView/useChatKeyboard/index.ios.ts src/components/KeyboardChatScrollView/useChatKeyboard/index.ts src/components/KeyboardChatScrollView/useChatKeyboard/__fixtures__/testUtils.ts src/components/KeyboardChatScrollView/useChatKeyboard/__tests__/index.ios.spec.ts src/components/KeyboardChatScrollView/useChatKeyboard/__tests__/freeze.android.spec.ts`
- `yarn typescript`

### DIFF
--- a/src/components/KeyboardChatScrollView/useChatKeyboard/index.ios.ts
+++ b/src/components/KeyboardChatScrollView/useChatKeyboard/index.ios.ts
@@ -1,4 +1,4 @@
-import { useSharedValue } from "react-native-reanimated";
+import { useAnimatedReaction, useSharedValue } from "react-native-reanimated";
 
 import { useKeyboardHandler } from "../../../hooks";
 import useScrollState from "../../hooks/useScrollState";
@@ -64,14 +64,12 @@ function useChatKeyboard(
       onStart: (e) => {
         "worklet";
 
-        if (freeze.value) {
-          return;
-        }
-
         if (e.height > 0) {
           // eslint-disable-next-line react-compiler/react-compiler
           targetKeyboardHeight.value = e.height;
         }
+
+        currentHeight.value = e.height;
 
         const effective = getEffectiveHeight(
           e.height,
@@ -108,6 +106,10 @@ function useChatKeyboard(
           blankSpace.value,
           effective + extraContentPadding.value,
         );
+
+        if (freeze.value) {
+          return;
+        }
 
         // persistent mode: when keyboard shrinks, clamp to valid range
         if (
@@ -229,9 +231,7 @@ function useChatKeyboard(
       onEnd: (e) => {
         "worklet";
 
-        if (freeze.value) {
-          return;
-        }
+        currentHeight.value = e.height;
 
         const effective = getEffectiveHeight(
           e.height,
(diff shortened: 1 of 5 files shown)


src/components/KeyboardChatScrollView/useChatKeyboard/index.ts
```diff
@@ -82,13 +86,10 @@ function useChatKeyboard(
...
         "worklet";
 
-        if (freeze.value) {
-          return;
-        }
-
         if (e.height > 0) {
           // eslint-disable-next-line react-compiler/react-compiler
           targetKeyboardHeight.value = e.height;
+          currentHeight.value = e.height;
```
kirillzyusko, line 92: Why do we need to update it inside `onStart`? We also update it inside `onMove`, so it would be strange to have following sequence of changes like 336, 0, 80, 120 ... 336?

src/components/KeyboardChatScrollView/useChatKeyboard/index.ios.ts
```diff
@@ -229,22 +231,38 @@ function useChatKeyboard(
...
+      if (!isFrozen && wasFrozen === true) {
+        padding.value = getEffectiveHeight(
+          currentHeight.value,
+          targetKeyboardHeight.value,
+          offset,
+        );
+      }
+    },
+    [offset],
+  );
```
kirillzyusko, line 264: This code is identical at the moment between iOS/Android? Would it make sense to move it into separate `useFrozenPadding` hook? Similar how code organized in `useExtraContentPadding`/`useEndVisible`? What do you think about it?

kirillzyusko:
Closing in favor of https://github.com/kirillzyusko/react-native-keyboard-controller/pull/1528

I will publish new changes in upcoming `1.21.14` version 🙌

Thank you for your contribution ❤️

Issue #1315, opened 2026-02-18
[Android] KeyboardAvoidingView retains stale keyboard height after app is backgrounded with keyboard visible


hosseintalebi:
**Describe the bug**
When the app is backgrounded while the keyboard is visible (e.g., on a text input screen), KeyboardAvoidingView with `behavior='height'` retains a stale keyboard height after the app returns to foreground. This causes screens that mount after foregrounding to render with their content compressed into the top ~50% of the screen, leaving blank space at the bottom.

**Code snippet**
  Minimal setup that reproduces the issue:
```tsx
  // App root
  <KeyboardProvider enabled>
    <NavigationContainer>
      <Stack.Navigator>
        <Stack.Screen name="Input" component={InputScreen} />
        <Stack.Screen name="Home" component={HomeScreen} />
      </Stack.Navigator>
    </NavigationContainer>
  </KeyboardProvider>

  // InputScreen — has a TextInput that opens the keyboard
  const InputScreen = () => (
    <KeyboardAvoidingView behavior="height" style={{ flex: 1 }}>
      <TextInput placeholder="Type here..." />
      <Button title="Go to Home" onPress={() => navigation.navigate('Home')} />
    </KeyboardAvoidingView>
  );

  // HomeScreen — content is compressed after backgrounding on InputScreen
  const HomeScreen = () => (
    <KeyboardAvoidingView behavior="height" style={{ flex: 1 }}>
      <Text>This content will be compressed into the top half of the screen</Text>
    </KeyboardAvoidingView>
  );
```

**To Reproduce**
  1. Open the app on Android and navigate to a screen with a TextInput wrapped in KeyboardAvoidingView with `behavior='height'`
  2. Tap the TextInput to open the keyboard
  3. Background the app
  4. Return to the app
  5. Navigate to a different screen that also uses KeyboardAvoidingView with `behavior='height'`
  6. Observe the content is compressed into the top ~50% of the screen

**Expected behavior**
The screen should render at full height. The keyboard controller's shared values should be reset when the keyboard is dismissed during backgrounding.

**Smartphone (please complete the following information):**

- Device: Pixel 7
- OS: Android 16
- RN version: 0.81.5
- RN architecture: new (Fabric)
- JS engine: Hermes
- Library version: 1.20.6

kirillzyusko:
Hey @hosseintalebi 

I used this code:

```tsx
import "react-native-gesture-handler";

[...]

PR #1526, opened 2026-07-02
fix: reset isInteractive when interactive gesture ends without animateToFinish

### LINKED ISSUE #1525: [Android] `isInteractive` never reset when an interactive gesture is cancelled — desync recovery permanently disabled, `KeyboardStickyView` strands until app kill
**Describe the bug**

`InteractiveKeyboardProvider.isInteractive` is a process-global flag set in `KeyboardAnimationController.startControlRequest()` / `insetBy()`, but cleared **only** in `animateToFinish()`. When an interactive dismiss gesture ends any other way — most commonly the user backgrounding the app mid-gesture (gesture-nav home swipe while the keyboard is being dragged) — the gesture ends via `cancel()` / `finish()` / `onCancelled → reset()`, and none of those paths clear the flag. It stays `true` for the life of the process.

The downstream effect is severe: `KeyboardAnimationCallback.onApplyWindowInsets` skips the *"detected desynchronized state — force updating it"* recovery whenever `isMoving = isTransitioning || InteractiveKeyboardProvider.isInteractive`. Once the flag is stranded, the library's only self-heal mechanism is **permanently disabled**. Any keyboard hide whose events are then lost while backgrounding (dropped view events, or `onEnd` emitting `keyboardDidShow` off a stale insets read during window teardown — see Additional context) leaves JS state stuck: `KeyboardController.isVisible()` returns `true` and the Reanimated `height` stays at the open value while the keyboard is actually closed. A `KeyboardStickyView` then renders its child one full keyboard-height above the bottom of the screen, with no keyboard visible, **until the app process is killed**.

We believe this is the root cause behind #1315 (closed as not reproducible). The apparent randomness is explained: the stranded state only *persists* when `isInteractive` (or `isTransitioning`) was already stranded beforehand — otherwise the desync recovery heals everything on the next foreground, which is why clean-environment repro attempts keep failing.

**Code snippet**

App-side setup is the canonical chat pattern from the docs:

```tsx
<KeyboardGestureArea interpolator="ios" style={{ flex: 1 }}>
  <MessagesList />
  <KeyboardStickyView offset={{ closed: 0, opened: bottom }}>
    <Composer />
  </KeyboardStickyView>
</KeyboardGestureArea>
```

**Repo for reproducing**

Not needed — the repro below is deterministic on a stock Android 16 emulator using only adb, in any app with the setup above. Happy to provide a minimal repo if useful.

**To Reproduce**

1. Focus the input → keyboard opens.
2. Strand the flag — begin an interactive dismiss and background mid-gesture without lifting the finger:

```bash
adb shell input motionevent DOWN 540 1300
adb shell input motionevent MOVE 540 1500
adb shell input motionevent MOVE 540 1700   # keyboard is now visibly dragging
adb shell input keyevent 3                  # HOME mid-drag → gesture cancelled, flag stranded
```

Observable in debug logcat from this point on: `DiffY: … true …` (the `isInteractive` field) on every subsequent animation, and *"detected desynchronized state"* lines **never appear again**.

3. Return to the app; open the keyboard cleanly (tap the input, let the animation finish) so `height = -H` and `isVisible = true`.
4. Press HOME with the keyboard open. On the emulator the IME hides with no insets animation on the app window → zero events reach JS, and with the recovery disabled nothing ever corrects the state. (To also suppress the emulator's IME-restore-on-return, which masks the visual: `adb shell am force-stop com.google.android.inputmethod.latin`.)
5. Reopen the app → the `KeyboardStickyView` child renders one keyboard-height above the bottom with the keyboard closed. It persists across further app switches; only killing the app resets it (native module re-init).

On real devices (Pixel-class, gesture nav) steps 4–5 happen naturally without adb: the user sends a message (input becomes non-editable → focus clears → IME hide starts) and backgrounds/app-switches mid-hide. That's how our production users hit it — several independent field reports, screenshot available.

**Expected behavior**

A cancelled interactive gesture resets `isInteractive`, so the desynchronized-state recovery keeps working and any stranded keyboard state self-corrects on the next foreground.

**Screenshots**

Field report and emulator repro show the identical state: composer floating mid-screen, keyboard-height gap below it, keyboard closed. Will attach both.

- Desktop OS: macOS 15
- Device: Pixel 10 Pro (field reports), Android 16 emulator (deterministic repro)
- OS: Android 16
- RN version: 0.85.3 (Expo SDK 56)
- RN architecture: new (Fabric, bridgeless)
- JS engine: Hermes
- Library version: 1.21.12 (verified unchanged on current `main`)

**Additional context**

Root-cause pointers:

- `android/src/main/java/com/reactnativekeyboardcontroller/interactive/KeyboardAnimationController.kt`: `isInteractive = true` in `startControlRequest()` (~L86) and `insetBy()` (~L141); cleared only in `animateToFinish()` (~L270). `cancel()` (~L207), `finish()` (~L220) and `reset()` (~L329) never clear it — including the `onCancelled` control-listener path.
- `android/src/main/java/com/reactnativekeyboardcontroller/listeners/KeyboardAnimationCallback.kt` (~L158): `val isMoving = isTransitioning || InteractiveKeyboardProvider.isInteractive` gates the desync recovery (~L180).

Proposed fix: clear `InteractiveKeyboardProvider.isInteractive` in `KeyboardAnimationController.reset()` — the single choke point reached by every gesture-end path. **PR with the fix: #1526**

Secondary observation from the same investigation (can file separately if you prefer): `onEnd` decides `keyboardDidHide` vs `keyboardDidShow` from a live `isKeyboardVisible()` insets read (`KeyboardAnimationCallback.kt` ~L336). During window teardown we consistently logged stale "IME visible" insets snapshots arriving 6–181ms after backgrounding, so an interrupted hide can end by emitting `keyboardDidShow` — leaving `KeyboardController.isVisible()` stuck `true` even without the interactive strand. Deriving the end-event from the animation's own target state instead of a live read would harden this path too.

idoyana:
## 📜 Description

Clear `InteractiveKeyboardProvider.isInteractive` in `KeyboardAnimationController.reset()`.

Today the flag is set in `startControlRequest()` / `insetBy()`, but cleared only in `animateToFinish()`. Gestures that end via `cancel()`, or via the control listener's `onFinished` / `onCancelled` paths, all funnel through `reset()` — and none of them clear the flag, so it strands `true` for the rest of the process lifetime. `reset()` is the single choke point every gesture-end path reaches, so clearing it there covers all of them (and is idempotent with the explicit clear in `animateToFinish()`).

## 💡 Motivation and Context

Fixes #1525.

A stranded `isInteractive` makes `KeyboardAnimationCallback` treat the keyboard as perpetually moving (`isMoving = isTransitioning || InteractiveKeyboardProvider.isInteractive`), which permanently disables the "detected desynchronized state" recovery in `onApplyWindowInsets`. After that, any keyboard hide whose events are lost while the app is backgrounded leaves `KeyboardController.isVisible()` stuck `true` and the Reanimated `height` at the open value — a `KeyboardStickyView` child then renders one keyboard-height above the bottom (keyboard closed) until the app is killed. Likely also the root cause behind the "not reproducible" #1315.

The easiest real-world strand: user drags the keyboard down (interactive dismiss) and backgrounds the app mid-gesture (gesture-nav home swipe) — the control request is cancelled, `animateToFinish()` never runs.

### Android

- reset `InteractiveKeyboardProvider.isInteractive` in `KeyboardAnimationController.reset()` so cancelled/aborted interactive gestures don't permanently disable the desynchronized-state recovery

Deterministic repro from #1525 on an Android 16 emulator (app using the docs' chat pattern — `KeyboardGestureArea` + `KeyboardStickyView`, RN 0.85.3, Fabric):

1. Open keyboard → start an interactive dismiss drag (`adb shell input motionevent DOWN/MOVE...`) → press HOME mid-drag (no UP).
2. Before the fix: every subsequent animation logs `DiffY: … true …` (stranded flag) and `detected desynchronized state` recovery lines never appear again; after backgrounding with the keyboard open, `KeyboardStickyView` content strands one keyboard-height above the bottom and never recovers (only app kill).
3. With the fix: the flag clears when the cancelled gesture funnels through `reset()`, the desync recovery fires on the next `onApplyWindowInsets`, and the stranded position self-corrects on foreground.

## 📝 Checklist

- [x] CI successfully passed

### DIFF
--- a/android/src/main/java/com/reactnativekeyboardcontroller/interactive/KeyboardAnimationController.kt
+++ b/android/src/main/java/com/reactnativekeyboardcontroller/interactive/KeyboardAnimationController.kt
@@ -337,6 +337,12 @@ internal class KeyboardAnimationController {
     currentSpringAnimation = null
 
     pendingRequestOnReady = null
+
+    // `animateToFinish` clears this flag before finishing the controller, but gestures that
+    // end via `cancel()`/`onFinished`/`onCancelled` never pass through it. A stranded `true`
+    // makes `KeyboardAnimationCallback` treat the keyboard as moving forever, which disables
+    // its desynchronized-state recovery for the rest of the process lifetime.
+    InteractiveKeyboardProvider.isInteractive = false
   }
 
   /**


kirillzyusko:
Hey @idoyana 👋 

Don't you mind to attach videos how it was before/after? Ideally if you can capture them with visible finger interaction (i. e. mouse cursor or touch cursor). I would like to reproduce it on my end to confirm that it actually fixes the issue 🙌

idoyana:
Hey @kirillzyusko 👋

Here are the before/after videos, recorded on an Android 16 emulator with touch indication enabled — the same scripted adb driver ran in both, so the interactions are identical.

📹 **Before (1.21.12):** [before-1.21.12.mp4](https://github.com/idoyana/react-native-keyboard-controller/releases/download/repro-1525-videos/before-1.21.12.mp4)
📹 **After (this PR):** [after-pr1526.mp4](https://github.com/idoyana/react-native-keyboard-controller/releases/download/repro-1525-videos/after-pr1526.mp4)

**Environment:** RN 0.85.3 (Fabric), library 1.21.12, Gboard, the docs' chat pattern — `KeyboardGestureArea interpolator="ios"` wrapping the message list + `KeyboardStickyView` for the composer.

**Exact steps (identical in both videos):**

1. Focus the input — keyboard opens.
2. Start an interactive-dismiss drag over the message list and press HOME **mid-drag** (finger never lifted). The gesture is cancelled → `onCancelled → reset()` runs, `animateToFinish()` never does. On 1.21.12 `isInteractive` strands `true`; with this PR `reset()` clears it.
3. Return to the app, close the keyboard (BACK), reopen it by tapping the input — a normal open, leaving `height = -336` / `isVisible = true`.
4. Press HOME with the keyboard open. On the emulator the IME hides with **no insets animation** on the app's window → zero events reach JS. (On real devices the same event loss happens when a user sends a message — the input becomes non-editable → focus clears → the IME hide starts — and they background mid-hide.)
5. `am force-stop com.google.android.inputmethod.latin` before returning — otherwise the emulator's IME-restore re-opens the keyboard on return and visually masks the stale state. Real users don't get that restore in the send flow, since focus was already cleared.
6. Return to the app.

**Before:** the composer renders one full keyboard-height above the bottom with the keyboard closed, because the desync recovery in `onApplyWindowInsets` is skipped (`isMoving` is stranded `true`). It persists across further app switches — only killing the app resets it. In debug logcat the `DiffY:` isInteractive field stays `true` forever and *"detected desynchronized state"* lines never appear again.

**After:** identical steps — the recovery fires on return (`detected desynchronized state - force updating it: 0.0` → `keyboardDidHide`) and the composer stays docked.

Happy to provide anything else that helps you reproduce it on your end 🙌

kirillzyusko (approved):

kirillzyusko:
I managed to reproduce it on physical Pixel 7 Pro API 36 using this:
- open "Interactive keyboard (android)" example;
- open keyboard;
- enter in terminal `sleep 10 && adb shell input keyevent 3`;
- start interactive gesture and hold your finger on the screen

After 10s app will be backgrounded. When you foreground it plain opening keyboard shows this (wrong `interactive` event instead of `move`):

Screenshot 2026-07-03 at 11 16 24

With fix we get this:

image

I also tested other scenarios such as:
- back gesture dismissal (correctly fires move event)
- start interactive dismissal and then move keyboard up - after that dismiss a keyboard via hide button (fires correct move events)
- dismiss keyboard interactively - tap to open a keyboard (correctly fires move event)
- reveal keyboard via gesture up (correctly fires interactive event)
- tested scenario with backgrounding app (correctly fires interactive and move events)

So far I haven't found any critical issues, so PR looks good for me ✅ 

Thank you for your contribution ❤️

idoyana:
Thank you for the fast review and for reproducing it on physical hardware — your hold-finger recipe is simpler than ours, we're adopting it 🙌 We're running this patch in production via `pnpm patch` and will drop it as soon as it ships in a release. Thanks for the great library ❤️

kirillzyusko:
Yes @idoyana I'll push a new version next week (most likely on Monday) 🤞

Issue #338, opened 2024-01-19
KeyboardAwareScrollView not working on first load of the app after fresh install in production iOS builds


domenicoprestia:
**Describe the bug**

First of all i want to thank you for the great work you've put in in this repo, it's a gamechanger.

I've a form that needs to be inside a KeyboardAwareScrollView, everytime i open the keyboard i've a button that remains sticky to the header of the keyboard (it works well on the first load after a fresh install).

The strange behaviour is that when i update or i install the application on my device (iPhone 12 Pro) the KeyboardAwareScrollView doesnt scroll how it should preventing the keyboard from covering the focused input.

If i reload the app it works fine.

**Code snippet**

  <View style={styles.container}>

    <View>

      <KeyboardAwareScrollView
        bottomOffset={88}
        ref={scrollRef}
        onScroll={scrollHandler}
        onContentSizeChange={callbackContentSizeChange}
        showsVerticalScrollIndicator={false}
        keyboardShouldPersistTaps={'handled'}
        keyboardDismissMode={'on-drag'}
      >
        <RecipeFormTitle
          step={step}
          setStep={setStep}
          paddingBottom={headerPaddingBottom}
          paddingTop={headerHeight}
          scrollValue={scrollValue}
        />
        {renderFormStep}
      </KeyboardAwareScrollView>
    </View>

    <RecipeFormFooter
      onContinue={callbackContinue}
      onBack={callbackBack}
      step={step}
      disabled={disabled} />

    <KeyboardFloatingCTA
      onPress={callbackPressFloatingCTA}
      text={textFloatingCTA}
      disabled={disabledFloatingCTA}
      icon={iconFLoatingCTA}
    />

    <ModalTagsConfirmation bottomSheetModalRef={modalTagsConfirmationRef} onContinue={callbackContinueForce} />
    <ModalStepsConfirmation bottomSheetModalRef={modalSingleStepConfirmationRef} onContinue={callbackContinueForce} />
    <ModalChallengeConfirmation bottomSheetModalRef={modalIngredientsChallengeConfirmationRef}onConfirm={callbackContinueForce}  />
    <ModalPhotoChallengeConfirmation bottomSheetModalRef={modalPhotoChallengeConfirmationRef} onConfirm={callbackContinueForce} />
    <ModalFailedImageUpload bottomSheetModalRef={modalFailedImageUploadRef} />
  </View>;

**Repo for reproducing**

The code is private, i cant share it with you sadly.

[...]

PR #1529, opened 2026-07-02
fix: KeyboardStickyView jumps on first keyboard appearance on iOS

invivek26:
## 📜 Description

Fixes `KeyboardStickyView` (and anything else driven by the `reanimated`/`animated` context values) teleporting to its final position on the **first keyboard appearance after app launch** on iOS, instead of animating together with the keyboard. Subsequent keyboard appearances in the same process are unaffected.

## 💡 Motivation and Context

On iOS the context values are driven by the **one-shot `onKeyboardMoveStart` event**: the final `height`/`progress` are written once at `keyboardWillShow` time, and the visual smoothness relies on that write being applied while UIKit's keyboard animation transaction is still in flight, so the transform change is implicitly animated by CoreAnimation with the keyboard's own curve.

On the very first keyboard show after app launch the event → worklet → style-application pipeline is cold (worklet/event-handler initialization, screen just mounted, busy main thread). The one-shot write then lands **outside** the animation transaction, so the sticky view snaps to the final position in a single frame while the keyboard is still animating up — the button visibly *leads* the keyboard. Once the pipeline is warm, the write lands inside the transaction and everything looks perfect, which is why only the first show per process is affected.

I verified the mechanism by logging inside `updateSharedValues`: on iOS the reanimated values receive exactly **one** update per transition (`height=308, progress=1.0` at will-show), while the native per-frame pipeline (`KeyboardMovementObserver.updateKeyboardFrame`, display link + CA animation extraction) emits a correct interpolated stream that was simply discarded by the platform filters — even on a cold first show.

The fix drives iOS from the per-frame `onKeyboardMove` stream plus `onKeyboardMoveEnd`, the same wiring Android uses. The per-frame values come from the native display link that samples/interpolates the keyboard's actual CA animation, so the sticky view follows the keyboard's real animated position without depending on implicit animation adoption. `onKeyboardMoveEnd` still guarantees the final value if any ticks are dropped.

Related cold-start report: #338 (first load after fresh install, iOS).

### JS

- iOS: drive `reanimated`/`animated` context values from per-frame `onKeyboardMove` + `onKeyboardMoveEnd` events instead of the one-shot `onKeyboardMoveStart` (fixes first-show jump of `KeyboardStickyView` and other consumers)

- Reproduced in a production app (Expo 54 / RN 0.81.4, new arch, iOS 26): fresh launch → navigate to a sign-in screen with an auto-focused input + `KeyboardStickyView` footer → footer jumps to final position while the keyboard animates (screen recording of the report matches the one-shot signature: button leads the keyboard).
- With this change, worklet logging shows a full per-frame ramp (~90 updates per show/hide cycle, `progress` 0.04 → 1.0 following the keyboard spring), and frame-by-frame analysis of a screen recording (30/60 fps) shows the footer rising in lockstep with the keyboard on the first show of a fresh process, and on all subsequent shows.
- `yarn typescript`, `yarn lint`, `yarn test` (all 202 library tests pass; the `docs/` and `FabricExample/` suite failures are pre-existing module-resolution issues in workspaces without installed deps).

## 📝 Checklist

- [ ] CI successfully passed
- [x] I added new mocks and corresponding unit-tests if library API was changed (no API change — event wiring only)

### DIFF
--- a/src/animated.tsx
+++ b/src/animated.tsx
@@ -158,17 +158,22 @@ export const KeyboardProvider = (props: KeyboardProviderProps) => {
       heightSV.value = -event.height;
     }
   };
+  // On iOS the values were historically driven by the one-shot
+  // `onKeyboardMoveStart` event: the final height was written once and the
+  // visual smoothness relied on that write being adopted by UIKit's in-flight
+  // keyboard animation transaction (implicit CoreAnimation animation). On the
+  // first keyboard show after app launch the event/worklet pipeline is cold and
+  // the write lands outside the animation transaction, so views driven by these
+  // values teleport to the final position while the keyboard is still animating.
+  // Driving iOS from the per-frame `onKeyboardMove` stream (interpolated
+  // natively from the keyboard's own CA animation) removes that dependency;
+  // `onKeyboardMoveEnd` guarantees the final value even if ticks are dropped.
   const keyboardHandler = useAnimatedKeyboardHandler(
     {
-      onKeyboardMoveStart: (event: NativeEvent) => {
-        "worklet";
-
-        updateSharedValues(event, ["ios"]);
-      },
       onKeyboardMove: (event: NativeEvent) => {
         "worklet";
 
-        updateSharedValues(event, ["android"]);
+        updateSharedValues(event, ["android", "ios"]);
       },
       onKeyboardMoveInteractive: (event: NativeEvent) => {
         "worklet";
@@ -178,7 +183,7 @@ export const KeyboardProvider = (props: KeyboardProviderProps) => {
       onKeyboardMoveEnd: (event: NativeEvent) => {
         "worklet";
 
-        updateSharedValues(event, ["android"]);
+        updateSharedValues(event, ["android", "ios"]);
       },
     },
     [],
@@ -223,10 +228,9 @@ export const KeyboardProvider = (props: KeyboardProviderProps) => {
         style={styles.container}
         // on*Reanimated prop must precede animated handlers to work correctly
         onKeyboardMoveReanimated={keyboardHandler}
-        onKeyboardMoveStart={OS === "ios" ? onKeyboardMove : undefined}
-        onKeyboardMove={OS === "android" ? onKeyboardMove : undefined}
+        onKeyboardMove={onKeyboardMove}
         onKeyboardMoveInteractive={onKeyboardMove}
-        onKeyboardMoveEnd={OS === "android" ? onKeyboardMove : undefined}
+        onKeyboardMoveEnd={onKeyboardMove}
         onFocusedInputLayoutChangedReanimated={inputLayoutHandler}
       >
         {children}


kirillzyusko:
Hey @invivek26 

I think correct fix for your use case is to understand why first time worklet runs with a delay.

This is the key feature of this library, that it drives animation with native performance (because uses CoreAnimation). With these changes we switch to Flutter behavior - update content position every frame. And while it works it'll consume more resources because animation gets driven by CPU instead of GPU.

So by default I would like to have highly optimized animations. If it doesn't work in one particular use-case, then we need to fix a root cause and don't try to hide a problem using a fallback method 🤷‍♂️ 

THe linked issue is incorrect: `KeyboardAwareScrollView` always uses `onMove` handler. The issue was in fact that those events were not firing in specific configurations 🤔 

I'll close this PR because I'm not going to merge...

invivek26:
Hey @kirillzyusko! Thanks for the quick look.

Just wanted to flag a couple of things that might change the picture:

1. **No new CPU work is introduced here.** The `CADisplayLink` in `KeyboardMovementObserver+Watcher` already fires every frame on iOS and emits `onKeyboardMove` events (it's how `KeyboardAwareScrollView` works today). This PR just stops discarding those events on the JS side — the `["android"]` filter becomes `["android", "ios"]`.

2. **It's not one particular use case.** Any screen that shows the keyboard early in the app lifecycle (sign-in with autofocus, onboarding, chat) hits this on every fresh launch. The root cause is that Reanimated's worklet initialization has non-deterministic latency on cold start, so the one-shot write misses UIKit's animation transaction — that's not really fixable from this library's side.

Happy to explore alternatives if you'd prefer a different approach (opt-in prop, worklet prewarming, etc) — just wanted to find a path forward since it's a pretty visible first-impression bug.

kirillzyusko:
> No new CPU work is introduced here. The CADisplayLink in KeyboardMovementObserver+Watcher already fires every frame on iOS and emits onKeyboardMove events (it's how KeyboardAwareScrollView works today). This PR just stops discarding those events on the JS side — the ["android"] filter becomes ["android", "ios"].

Yes, exactly. `KeyboardAwareScrollView` uses `useKeyboardHandler` and intentionally uses frame-by-frame driven animation (mostly because it is running side effects, such as calling `scrollTo` each frame to interpolate scrolling).

With this PR you make **ALL** components to be driven frame-by-frame (`KeyboardStickyView`, `KeyboardToolbar`, upcoming `KeyboardEffects`). So you literally migrate animation from GPU to CPU 🤷‍♂️ So you **DO** introduce new CPU work.

> It's not one particular use case. Any screen that shows the keyboard early in the app lifecycle (sign-in with autofocus, onboarding, chat) hits this on every fresh launch. The root cause is that Reanimated's worklet initialization has non-deterministic latency on cold start, so the one-shot write misses UIKit's animation transaction — that's not really fixable from this library's side.

Which library version do you use? According to the description of the problem I assume it is `1.21.7+`? In `1.21.7` I [migrated](https://github.com/kirillzyusko/react-native-keyboard-controller/commit/21a939f30de8b73e68410bcde169988ad1f87c9c) `KeyboardStickyView` back to `reanimated`. Which architecture do you use? New or old? If new - then react commit may produce [jumpy](https://kirillzyusko.github.io/react-native-keyboard-controller/docs/api/components/keyboard-chat-scroll-view#missing-animations-on-ios-new-arch-only) animation (don't pay attention that it's `KeyboardChatScrollView` - the same problem is applicable to many components - `KeyboardAvoidingView` and `KeyboardStickyView` too). Have you tried to enable [this](https://docs.swmansion.com/react-native-reanimated/docs/guides/feature-flags/#disable_commit_pausing_mechanism) feature flag in reanimated? Does it solve the problem?

Looking forward to your reply. If you can give me answers on my questions we can plan next steps on how to fix this problem 🤞

Issue #1411, opened 2026-03-27
KeyboardToolbar persists after keyboard dismiss + KeyboardAwareScrollView doesn't scroll to focused input (Expo 55 / RN 0.83 / Fabric)


imckain:
**Describe the bug**
After upgrading to Expo SDK 55 (React Native 0.83.2, Fabric/New Architecture, react-native-reanimated 4.2.1), two components are broken:

1. `KeyboardToolbar` remains visible on screen after the keyboard is dismissed. The `translateY` transform on the underlying `KeyboardStickyView` gets stuck at the "keyboard open" position and never resets.
2. `KeyboardAwareScrollView` does not auto-scroll to the focused `TextInput` when the keyboard appears. The content stays in place as if no keyboard event was received.

Additionally, this warning appears in the console logs:
```
Can not attach worklet handlers for `react-native-keyboard-controller` because view tag can not be resolved. Be sure that `KeyboardProvider` is fully mounted before registering handlers.
```

This suggests the native `KeyboardControllerView` isn't resolving its view tag before worklet handlers attempt to attach, meaning keyboard events may not be propagating at all.

**Code snippet**
```
// KeyboardToolbar usage (currently returning null as workaround)
import { KeyboardToolbar as RNKeyboardToolbar } from 'react-native-keyboard-controller';

// Renders outside the scroll view, sibling to ModalScreenWrapper
<>
  <KeyboardAwareScrollView>{/* form content */}</KeyboardAwareScrollView>
  <KeyboardToolbar />
</>
```

**Repo for reproducing**
I will add if I can get time

**To Reproduce**

- Wrap app in `KeyboardProvider`
- Render `KeyboardToolbar` as a sibling to a scroll view containing TextInput fields
- Tap a `TextInput` to open the keyboard
- Dismiss the keyboard
- Observe: toolbar remains visible, scroll view never adjusted

**Expected behavior**

- `KeyboardToolbar` should slide down and hide when the keyboard dismisses
- `KeyboardAwareScrollView` should auto-scroll to keep the focused input visible above the keyboard

**Smartphone (please complete the following information):**

- Desktop OS: [MacOS 15.7.5]
- Device: [iPhone 16, iOS sim]
- OS: [iOS 26.x]
- RN version: [83.2]
- RN architecture: [fabric]
- JS engine: [Hermes]
- Library version: [1.21.2]

**Additional context**

[...]

Issue #1316, opened 2026-02-19
[Android]: Content not focused when keyboard opens


iqraKhaliq:
**Describe the bug**
On Android, when the keyboard opens and a user focuses on an input field near the bottom of the screen, the content does not automatically scroll into view. The focused input becomes partially hidden behind the keyboard. The KeyboardAwareScrollView is not properly adjusting the scroll position.

**Code snippet**
```tsx
<KeyboardAwareScrollView
          bounces={false}
          screenReaderFocusable
          disableScrollOnKeyboardHide
          style={styles.keyboardContainer}
          bottomOffset={extraScrollHeight}
          automaticallyAdjustKeyboardInsets
          keyboardShouldPersistTaps={'handled'}
          showsVerticalScrollIndicator={showsVerticalScrollIndicator}
          contentContainerStyle={{ flexGrow: 1 }}
        >
          {renderScrollableContent()}
        </KeyboardAwareScrollView>
````

**Steps to reproduce the behavior**
- Launch the app on Android
- Tap on an input field near the bottom (e.g., Confirm Password)
- Observe when the keyboard opens

**Expected behavior**
The screen should automatically scroll so that the focused input field remains fully visible above the keyboard like on iOS

**Actual behavior**
The keyboard opens but the scroll view does not adjust correctly. The focused input is partially hidden behind the keyboard.

**Demo Videos**

https://github.com/user-attachments/assets/eb51c7f8-a161-4b86-ac2c-fc70ca08df98

https://github.com/user-attachments/assets/151f1e3f-0089-457d-8816-221bb7bfd26e

**Smartphone (please complete the following information):**
- Desktop OS: macOS 26.2
- Device: Android Emulator (Medium Phone API 36)
- OS: Android 16 (API 36)
- RN version: 0.83.1
- RN architecture: New Architecture
- JS engine: Hermes

**Additional context**
- Issue occurs only on Android.
- Using KeyboardAwareScrollView inside a form layout.
- The problem happens consistently when focusing lower input fields.
- automaticallyAdjustKeyboardInsets and bottomOffset do not resolve the issue.

kirillzyusko:
Similar issue has been reported here, I believe: https://github.com/kirillzyusko/react-native-keyboard-controller/issues/1300

@iqraKhaliq which input component do you use for 6-digit code field?

iqraKhaliq:
@kirillzyusko  The issue persists on Android even when OTP fields are not used.

[...]

PR #1537, opened 2026-07-09
fix(android): resolve enclosing scroll view when useNestedScrollViewAndroid is enabled

debitan:
## 📜 Description

`EditText.parentScrollViewTarget` only matches `ReactScrollView`, but when the `useNestedScrollViewAndroid` feature flag is enabled RN renders the vertical scroll view as `ReactNestedScrollView`. The walk therefore never finds the enclosing scroll view and returns `-1`, so `KeyboardAwareScrollView`'s `maybeScroll` bails on the `parentScrollViewTarget !== scrollViewTarget` guard and the focused input is never scrolled above the keyboard.

`ReactNestedScrollView` is package-private, so it cannot be referenced by type from outside `com.facebook.react.views.scroll`. This matches its public superclass `androidx.core.widget.NestedScrollView` instead (which `ReactNestedScrollView` extends).

## 💡 Motivation and Context

Fixes the Android auto-scroll failure discussed in #1411 for apps that enable `useNestedScrollViewAndroid` (commonly turned on for nested scrolling inside native bottom sheets). With the flag on, `KeyboardAwareScrollView` stopped scrolling focused inputs into view on Android.

### Android

- Also match `NestedScrollView` (the public superclass of `ReactNestedScrollView`) in `parentScrollViewTarget`, so auto-scroll works when the `useNestedScrollViewAndroid` feature flag is enabled.

Tested in a React Native 0.85 (New Architecture) app on a physical Pixel and an emulator with `useNestedScrollViewAndroid` enabled. Before: inputs near the bottom of a `KeyboardAwareScrollView` were covered by the keyboard, and `parentScrollViewTarget` logged as `-1`. After: focused inputs scroll above the keyboard as expected. iOS is unaffected.

## 📝 Checklist

- [ ] CI successfully passed

### DIFF
--- a/android/src/main/java/com/reactnativekeyboardcontroller/extensions/EditText.kt
+++ b/android/src/main/java/com/reactnativekeyboardcontroller/extensions/EditText.kt
@@ -8,6 +8,7 @@ import android.view.Gravity
 import android.view.View
 import android.view.ViewTreeObserver.OnPreDrawListener
 import android.widget.EditText
+import androidx.core.widget.NestedScrollView
 import com.facebook.react.views.scroll.ReactScrollView
 import com.facebook.react.views.textinput.ReactEditText
 import com.reactnativekeyboardcontroller.log.Logger
@@ -105,7 +106,9 @@ val EditText.parentScrollViewTarget: Int
     while (currentView != null) {
       val parentView = currentView.parent as? View
 
-      if (parentView is ReactScrollView && parentView.scrollEnabled) {
+      if ((parentView is ReactScrollView && parentView.scrollEnabled) ||
+        parentView is NestedScrollView
+      ) {
         // If the parent is a vertical, scrollable ScrollView - return its id
         return parentView.id
       }


android/src/main/java/com/reactnativekeyboardcontroller/extensions/EditText.kt
```diff
@@ -105,7 +106,9 @@ val EditText.parentScrollViewTarget: Int
     while (currentView != null) {
       val parentView = currentView.parent as? View
 
-      if (parentView is ReactScrollView && parentView.scrollEnabled) {
+      if ((parentView is ReactScrollView && parentView.scrollEnabled) ||
+        parentView is NestedScrollView
+      ) {
```
kirillzyusko, line 111: While I like these changes, we miss one important thing - if `ScrollView` will be with disabled scroll then this code still return it as valid scrollable container.

Do you think we can add extension that can handle both cases?

```kt
import android.view.View
import com.facebook.react.views.scroll.ReactScrollView
import java.lang.reflect.Field
import java.lang.reflect.Method

internal fun View.isEnabledReactScrollView(): Boolean {
  if (this is ReactScrollView) {
    return scrollEnabled
  }

  if (ReactNestedScrollViewCompat.isInstance(this)) {
    return ReactNestedScrollViewCompat.isScrollEnabled(this)
  }

  return false
}

private object ReactNestedScrollViewCompat {
  private const val CLASS_NAME =
    "com.facebook.react.views.scroll.ReactNestedScrollView"

  private val clazz: Class<*>? by lazy(LazyThreadSafetyMode.PUBLICATION) {
    runCatching {
      Class.forName(
        CLASS_NAME,
        false,
        ReactScrollView::class.java.classLoader,
      )
    }.getOrNull()
  }

  private val scrollEnabledGetter: Method? by lazy(LazyThreadSafetyMode.PUBLICATION) {
    val c = clazz ?: return@lazy null

    runCatching {
      c.getMethod("getScrollEnabled")
    }.getOrNull()
      ?: runCatching {
        c.getMethod("isScrollEnabled")
      }.getOrNull()
  }

  private val scrollEnabledField: Field? by lazy(LazyThreadSafetyMode.PUBLICATION) {
    val c = clazz ?: return@lazy null

    runCatching {
      c.getDeclaredField("mScrollEnabled").apply {
        isAccessible = true
      }
    }.getOrNull()
  }

  fun isInstance(view: View): Boolean {
    return clazz?.isInstance(view) == true
  }

  fun isScrollEnabled(view: View): Boolean {
    scrollEnabledGetter?.let { getter ->
      return runCatching {
        getter.invoke(view) as Boolean
      }.getOrDefault(false)
    }

    scrollEnabledField?.let { field ->
      return runCatching {
        field.getBoolean(view)
      }.getOrDefault(false)
    }

    return false
  }
}
```

And use it like:

```kt
if (parentView.isEnabledReactScrollView()) {
```

I'm fine if you can simpler/better code. My main concern is that we need to check `scrollEnabled` property too.

Ideally would be also to check that it returns `false` with disabled `ScrollView` + check any possible performance impact. The reflection is an expensive thing and calling it for each  view might be expensive 🤞 

Let me know what do you think about it!

kirillzyusko:
Could be also related to https://github.com/kirillzyusko/react-native-keyboard-controller/issues/1316

kirillzyusko:
Closed in favour of https://github.com/kirillzyusko/react-native-keyboard-controller/pull/1639

PR #1546, opened 2026-07-13
fix: reset keyboard values when disabled

### LINKED ISSUE #1416: useReanimatedKeyboardAnimation animated values are not reset after KeyboardController enable/disable
## Describe the bug
When `KeyboardController` is disabled via `setEnabled(false)`, animated values from `useReanimatedKeyboardAnimation` stop updating.

If the keyboard has not fully finished closing before the controller is disabled, the hook keeps the last keyboard height value. After enabling the controller again, `useReanimatedKeyboardAnimation` still returns those stale values even though the keyboard is already fully closed.

This causes the hook state to be out of sync with the real keyboard state.

## Code snippet
```tsx
const keyboardController = useKeyboardController();

keyboardController.setEnabled(false);
```

I am using `useReanimatedKeyboardAnimation`, and after disabling the controller, its animated values are no longer updated or reset.

## Repo for reproducing
I do not have a reproduction repository yet.

## To Reproduce

1. Use `useReanimatedKeyboardAnimation`
2. Open the keyboard
3. Disable `KeyboardController` before the keyboard fully finishes closing:
   ```tsx
   const keyboardController = useKeyboardController();
   keyboardController.setEnabled(false);
   ```
4. Wait until the keyboard is fully closed
5. Enable `KeyboardController` again
6. Check animated values from `useReanimatedKeyboardAnimation`
7. Observe that they still contain stale keyboard height / previous values

## Expected behavior
When `KeyboardController` is disabled, animated values inside `useReanimatedKeyboardAnimation` should be reset, so that after enabling the controller again they are either:

- `0` when the keyboard is fully closed, or
- the current actual values that reflect the real keyboard position

The hook should not keep stale values from the previous keyboard state.

## Screenshots
Not attached.

## Smartphone (please complete the following information):

- Desktop OS: macOS
- Device: Google Pixel 8a
- OS: Android
- RN version: 0.81.5
- RN architecture: [old]
- JS engine: Hermes
- Library version: 1.19.1

## Additional context
The issue happens specifically when the controller is disabled while the keyboard transition is still in progress. In that case, `useReanimatedKeyboardAnimation` keeps the last known animated values and does not synchronize them with the actual closed keyboard state after re-enabling the controller.

MarshallBear1:
## 📜 Description

Reset Animated and Reanimated keyboard values, plus the cached focused-input layout, when the controller is disabled mid-transition.

Fixes #1416.

### JS

- clear stale keyboard animation state when disabled
- add a regression test for disabling mid-transition

- focused Jest regression suite
- TypeScript and ESLint
- disabled during keyboard focus, re-enabled, and reopened the keyboard on iPhone 17 / iOS 26.5 Simulator

## 📝 Checklist

- [ ] CI successfully passed
- [x] Added regression tests


src/animated.tsx
```diff
@@ -96,6 +97,18 @@ export const KeyboardProvider = (props: KeyboardProviderProps) => {
   const layout = useSharedValue<FocusedInputLayoutChangedEvent | null>(null);
   const setKeyboardHandlers = useEventHandlerRegistration(viewRef);
   const setInputHandlers = useEventHandlerRegistration(viewRef);
+
+  useLayoutEffect(() => {
+    if (!enabled) {
+      progress.setValue(0);
+      height.setValue(0);
```
kirillzyusko, line 104: Does `setValue` really works if the value is driven by native driver?..

PR #1570, opened 2026-07-28
fix: use propagation view `surfaceId` for focused input events inside `Modal`

### LINKED ISSUE #1569: Android Fabric focused-input events use wrong modal surface ID
**Describe the bug**

On Android with Fabric, focused-input events inside a React Native `Modal` use the modal window/root view to resolve their surface ID, but use the React propagation view as their event target.

`ModalAttachedWatcher` intentionally passes two different views:

```kotlin
FocusedInputObserver(rootView, eventPropagationView, context)
```

`FocusedInputObserver` currently combines values from different trees:

```kotlin
private val surfaceId = UIManagerHelper.getSurfaceId(view)

FocusedInputTextChangedEvent(
  surfaceId,
  eventPropagationView.id,
  text = text,
)
```

For a modal window/root view, `UIManagerHelper.getSurfaceId(view)` resolves to `-1`. The focused-input layout, text, and selection events should use the surface that owns `eventPropagationView`.

`KeyboardAnimationCallback` already resolves its surface ID from `eventPropagationView`, so focused-input events currently behave differently from keyboard animation events on the same modal path.

**Code snippet**

The proposed fix is:

```kotlin
private val surfaceId = UIManagerHelper.getSurfaceId(eventPropagationView)
```

**Repo for reproducing**

The repository's existing `example` app already contains a React Native `Modal` example. The regression can be demonstrated by adding a focused-input handler to that modal and asserting its text/selection events under Fabric.

**To Reproduce**

1. Run the Android Fabric example.
2. Open the existing Modal example.
3. Focus a `TextInput` rendered inside the modal.
4. Type or change the selection while observing focused-input text/selection events.
5. Inspect the dispatched event: its target belongs to `eventPropagationView`, but its surface ID is `-1` because it was resolved from the modal root view.

**Expected behavior**

Focused-input layout, text, and selection events use the surface ID associated with `eventPropagationView`, matching their event target and `KeyboardAnimationCallback`.

**Screenshots**

Not applicable; this is an event-routing mismatch.

- Desktop OS: macOS
- Device: Android
- OS: Android 16
- RN version: 0.86.0
- RN architecture: Fabric
- JS engine: Hermes
- Library version: 1.21.9 and current `main`

**Additional context**

This appears related to the modal support introduced in #466, which split the watched modal root from the React event propagation view. The normal non-modal path passes the same object for both parameters and is unaffected.

JoaoPauloCMarra:
## 📜 Description

Resolve Android focused-input event surface IDs from the React event propagation view instead of the watched modal window root.

## 💡 Motivation and Context

`ModalAttachedWatcher` observes the modal window root while dispatching focused-input events through a separate React view. Resolving the surface ID from the window root returns `-1` under Fabric, producing a surface/target mismatch for layout, text, and selection events.

The non-modal path is unchanged because both constructor arguments reference the same React view.

Fixes #1569.

### Android

- Fixed focused-input event routing inside React Native modals under Fabric

- Android debug assembly with React Native 0.76.2
- Android lint, detekt, and ktlint
- TypeScript, ESLint, and Prettier checks

## 📝 Checklist

- [x] CI successfully passed
- [x] Library API is unchanged; Android regression coverage was added


android/src/main/java/com/reactnativekeyboardcontroller/listeners/FocusedInputObserver.kt
```diff
@@ -43,7 +43,7 @@ class FocusedInputObserver(
   private val context: ThemedReactContext?,
 ) {
   // constructor variables
-  private val surfaceId = UIManagerHelper.getSurfaceId(view)
+  internal val surfaceId = UIManagerHelper.getSurfaceId(eventPropagationView)
```
kirillzyusko: Could we change it to private again? 🙏

android/src/test/java/com/reactnativekeyboardcontroller/listeners/FocusedInputObserverTest.kt
kirillzyusko, line 1: Can we remove this file? I think it's over-engineering to cover it with unit tests 😅

kirillzyusko (commented):
Thank you for PR! Once you address comments we can merge it 🙌

PR #1596, opened 2026-08-10
fix: reanimated 3.0 compatibility

### LINKED ISSUE #1585: [Android] compileReleaseKotlin fails on RN < 0.74 since 1.21.0: Unresolved reference ClippingScrollViewDecoratorViewManager (wrong package declaration)
**Describe the bug**

Android release builds fail on React Native < 0.74 (old architecture) since v1.21.0:

```
> Task :react-native-keyboard-controller:compileReleaseKotlin FAILED

e: .../android/src/turbo/java/com/reactnativekeyboardcontroller/KeyboardControllerPackage.kt:63:7
   Unresolved reference: ClippingScrollViewDecoratorViewManager
```

**Root cause**

`ClippingScrollViewDecoratorViewManager.kt` in both the `paper` and `fabric` source sets declares:

```kotlin
package java.com.reactnativekeyboardcontroller
```

instead of `package com.reactnativekeyboardcontroller` — it looks like the `java/` source-root folder got baked into the package name when the file was created. Every other file in the library is in `com.reactnativekeyboardcontroller`.

The `base` variant of `KeyboardControllerPackage.kt` (used when RN >= 0.74) compensates with an explicit import, so most users never hit this:

```kotlin
import java.com.reactnativekeyboardcontroller.ClippingScrollViewDecoratorViewManager
```

But the `turbo` variant (compiled when `shouldUseBaseReactPackage()` is false, i.e. RN < 0.74) references the class unqualified with no import, so it only resolves if the class is in `com.reactnativekeyboardcontroller` — which it isn't. Hence the compile failure on every RN < 0.74 build since the component was introduced in 1.21.0. Still present on `main` as of today.

**Suggested fix**

Rename the package in `android/src/paper/.../ClippingScrollViewDecoratorViewManager.kt` and `android/src/fabric/.../ClippingScrollViewDecoratorViewManager.kt` to `com.reactnativekeyboardcontroller`, and drop the `java.com.` import from the `base` `KeyboardControllerPackage.kt`. (As a side benefit, that moves the class out of the `java.*` namespace, which is prohibited on the JVM ClassLoader and only works on Android because classes load from dex.)

**To Reproduce**

1. RN 0.72.x app, old architecture (`newArchEnabled=false`)
2. Install `react-native-keyboard-controller` >= 1.21.0
3. `cd android && ./gradlew :react-native-keyboard-controller:compileReleaseKotlin`
4. Compilation fails with the unresolved reference above

**Expected behavior**

The library compiles on all RN versions it supports.

**Workaround**

patch-package adding the same import line the `base` variant uses to `android/src/turbo/.../KeyboardControllerPackage.kt`:

```diff
 import com.reactnativekeyboardcontroller.modules.statusbar.StatusBarManagerCompatModuleImpl
+import java.com.reactnativekeyboardcontroller.ClippingScrollViewDecoratorViewManager
```

**Environment**

- Desktop OS: macOS 15 / ubuntu-latest CI
- RN version: 0.72.3
- RN architecture: old/paper
- JS engine: Hermes
- Library version: 1.22.3 (affects 1.21.0 through 1.22.3 and current `main`)

kirillzyusko:
## 📜 Description

Fixed package compatibility for reanimated version lower than 3.7.0

## 💡 Motivation and Context

On Reanimated `3.7.0-` the `useEvent` handler returns `ref` (with `current` property). After reanimated 3.8.0 it returns an object that contains `workletEventHandler`. Somehow in https://github.com/kirillzyusko/react-native-keyboard-controller/pull/1209 I thought that it returns the event handler directly (not in ref). So I declared the object as object-or-direct-handler.

In https://github.com/kirillzyusko/react-native-keyboard-controller/issues/1585#issuecomment-5223569485 I realized that I made a mistake, so in this PR I eliminated a dead code and fixed compatibility.

Follow up for https://github.com/kirillzyusko/react-native-keyboard-controller/pull/1209 (something that had to be implemented there from the beginning).

Closes https://github.com/kirillzyusko/react-native-keyboard-controller/issues/1585#issuecomment-5223569485

### JS

- rename `WorkletHandlerOrWorkletHandlerObject` to `WorkletHandlerContainer`;
- replace `currentHandler.*` with `currentHandler.current.*`;
- updated titles of unit tests;

Tested via this PR (e2e tests).

image

## 📝 Checklist

- [x] CI successfully passed

### DIFF
--- a/src/__tests__/internal.spec.ts
+++ b/src/__tests__/internal.spec.ts
@@ -50,7 +50,7 @@ beforeEach(() => {
 });
 
 describe("useEventHandlerRegistration", () => {
-  it("should support a handler wrapped in workletEventHandler", () => {
+  it("should support the handler shape from Reanimated 3.8 and newer", () => {
     const viewTagRef = { current: VIEW };
     const workletEventHandler = createWorkletHandler();
     const register = renderRegistration(viewTagRef);
@@ -70,12 +70,14 @@ describe("useEventHandlerRegistration", () => {
     );
   });
 
-  it("should support a direct worklet handler", () => {
+  it("should support the ref handler shape from Reanimated 3.0 through 3.7", () => {
     const viewTagRef = { current: VIEW };
     const workletEventHandler = createWorkletHandler();
     const register = renderRegistration(viewTagRef);
 
-    const cleanup = register(workletEventHandler as unknown as EventHandler);
+    const cleanup = register({
+      current: workletEventHandler,
+    } as unknown as EventHandler);
 
     expect(workletEventHandler.registerForEvents).toHaveBeenCalledWith(
       VIEW_TAG,
--- a/src/internal.ts
+++ b/src/internal.ts
@@ -12,8 +12,10 @@ type WorkletHandler = {
   unregisterFromEvents: (viewTag: number) => void;
 };
 
-type WorkletHandlerOrWorkletHandlerObject =
-  | WorkletHandler
+type WorkletHandlerContainer =
+  | {
+      current: WorkletHandler;
+    }
   | {
       workletEventHandler: WorkletHandler;
     };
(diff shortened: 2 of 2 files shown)


kirillzyusko:
Hey @ahulterstrom 👋 

Could you please confirm that these changes fix a problem 2 described in https://github.com/kirillzyusko/react-native-keyboard-controller/issues/1585#issuecomment-5223569485? 🙏

PR #1611, opened 2026-08-27
fix: omit Reanimated dependencies on native

### LINKED ISSUE #1621: [Bug]: react-native-keyboard-controller triggers Reanimated 4.6.0 native dependency warning
**Describe the bug**
When using the keyboard handling utilities or hooks from this library on native platforms (iOS/Android), React Native Reanimated logs a warning to the console stating that `dependencies should only be used in web implementation.` 

This issue stems from an internal call to `useHandler` in `reanimated.native.ts` triggered during keyboard updates (`updateSharedValues`). A similar issue was recently discussed in `react-native-bottom-sheet` repository [here](https://github.com/gorhom/react-native-bottom-sheet/issues/2758).

**Code snippet**
The warning originates from internal hooks configuration mapping to Reanimated handler files:

```
node_modules\react-native-keyboard-controller\src\reanimated.native.ts
node_modules\react-native-keyboard-controller\src\animated.tsx
```

Logs:

```
 WARN  [Reanimated] dependencies should only be used in web implementation.

Call Stack
  logToConsole (node_modules\react-native-reanimated\src\common\logger.ts)
  handleLog (node_modules\react-native-reanimated\src\common\logger.ts)
  reactNativeReanimated_loggerTs5 (http:\10.9.9.16:8081\node_modules\expo\AppEntry.bundle)
  useHandler (http:\10.9.9.16:8081\node_modules\expo\AppEntry.bundle)
  <global> (node_modules\react-native-keyboard-controller\src\reanimated.native.ts)
  updateSharedValues (node_modules\react-native-keyboard-controller\src\animated.tsx)
  callComponent.react_stack_bottom_frame (node_modules\react-native\Libraries\Renderer\implementations\ReactFabric-dev.js)
  renderWithHooks (node_modules\react-native\Libraries\Renderer\implementations\ReactFabric-dev.js)
  updateFunctionComponent (node_modules\react-native\Libraries\Renderer\implementations\ReactFabric-dev.js)
  beginWork (node_modules\react-native\Libraries\Renderer\implementations\ReactFabric-dev.js)
  runWithFiberInDEV (node_modules\react-native\Libraries\Renderer\implementations\ReactFabric-dev.js)
  performUnitOfWork (node_modules\react-native\Libraries\Renderer\implementations\ReactFabric-dev.js)
  renderRootSync (node_modules\react-native\Libraries\Renderer\implementations\ReactFabric-dev.js)
  renderRootSync (node_modules\react-native\Libraries\Renderer\implementations\ReactFabric-dev.js)
  performWorkOnRoot (node_modules\react-native\Libraries\Renderer\implementations\ReactFabric-dev.js)
  performSyncWorkOnRoot (node_modules\react-native\Libraries\Renderer\implementations\ReactFabric-dev.js)
  flushSyncWorkAcrossRoots_impl (node_modules\react-native\Libraries\Renderer\implementations\ReactFabric-dev.js)
  markRootSuspended (node_modules\react-native\Libraries\Renderer\implementations\ReactFabric-dev.js)
  scheduleRefresh (node_modules\react-native\Libraries\Renderer\implementations\ReactFabric-dev.js)
  apply (<native>)
  renderer.scheduleRefresh (node_modules\react-devtools-core\dist\backend.js)
  mountedRootsSnapshot.forEach$argument_0 (node_modules\react-refresh\cjs\react-refresh-runtime.development.js)
  forEach (<native>)
  failedRootsSnapshot.forEach$argument_0 (node_modules\react-refresh\cjs\react-refresh-runtime.development.js)
  <global> (node_modules\react-native\Libraries\Core\setUpReactRefresh.js)
  setTimeout$argument_0 (node_modules\@expo\cli\build\metro-require\require.js)
```

**Repo for reproducing**
Easy to repro - Just use reanimated `4.6.0` with this library. 

**To Reproduce**

1. Mount `KeyboardProvider` or use keyboard hooks like `useAnimatedKeyboardHandler` / `useReanimatedKeyboardAnimation` inside a native screen.
2. Focus on a `TextInput` to pull up the keyboard, or trigger a layout refresh.
3. Observe the `WARN  [Reanimated] dependencies should only be used in web implementation.` outputted multiple times in the terminal/bundler logs.

**Expected behavior**
The hooks should execute cleanly without outputting console warnings on native platforms (iOS and Android).

- Desktop OS: Windows 11
- Device: iPhone XR
- OS: 18.7.11
- RN version: 0.83.10
- RN architecture: New
- JS engine: default
- Library version: 4.6.0 (does not occur in older versions)

**Additional context**
This message usually implies that a dependency array or second argument is accidentally being fed into a Reanimated custom hook wrapper (`useHandler`) while running on native, which Reanimated strictly warns against outside of `react-native-web`.

huextrat:
## 📜 Description

Route Reanimated's `useHandler` through the existing platform-specific `reanimated.native.ts` and `reanimated.ts` modules. The native wrapper omits dependency arrays, while the Web implementation re-exports Reanimated's original hook and continues to receive them.

The existing public hook signatures remain unchanged.

## 💡 Motivation and Context

Reanimated 4.6 warns whenever dependencies are supplied to native hooks because dependency arrays are only used on Web without the Babel plugin. `KeyboardProvider` and the public keyboard event hooks were forwarding dependency arrays to the native implementation, which could produce warnings during application startup and handler registration.

Native worklet closures already determine when handlers need to be rebuilt. Keeping this behavior behind the platform-specific module also avoids inline platform checks and preserves Web behavior for future support.

Closes https://github.com/kirillzyusko/react-native-keyboard-controller/issues/1621

### JS

- Avoid passing Web-only dependency arrays to native Reanimated handlers.
- Preserve dependency arrays on Web through a platform-specific `useHandler` implementation.

- `yarn typescript`
- `yarn lint`
- `yarn test src --runInBand --no-watchman` — 27 suites and 214 tests passed
- Reproduced the warnings with Reanimated 4.6 before the change in a React Native application using `KeyboardProvider`.

Not applicable; this change removes development console warnings and has no visual impact.

## 📝 Checklist

- [ ] CI successfully passed


kirillzyusko:
Hey @huextrat 

Could you please tell me, which warning you've seen exactly? And maybe who triggers a warning?

I quickly checked [this](https://github.com/software-mansion/react-native-reanimated/blob/main/packages/react-native-reanimated/src/hook/useHandler.ts) code, but I don't see any warnings been thrown in this hook?

huextrat:
> Hey @huextrat
> 
> Could you please tell me, which warning you've seen exactly? And maybe who triggers a warning?
> 
> I quickly checked [this](https://github.com/software-mansion/react-native-reanimated/blob/main/packages/react-native-reanimated/src/hook/useHandler.ts) code, but I don't see any warnings been thrown in this hook?

Hey @kirillzyusko

Sure, the exact warning is:
`[Reanimated] dependencies should only be used in web implementation.`

My setup is:
- react-native-keyboard-controller@1.22.4
- react-native-reanimated@4.6.0
- React Native 0.86.3, New Architecture, iOS development build

The important detail is that the file you linked is useHandler.ts. On React Native, Metro resolves [useHandler.native.ts](https://github.com/software-mansion/react-native-reanimated/blob/4.6.0/packages/react-native-reanimated/src/hook/useHandler.native.ts#L33-L35). In Reanimated 4.6.0, that implementation warns whenever the dependencies argument is not undefined.

The call chain I found is:
1. [KeyboardProvider](https://github.com/kirillzyusko/react-native-keyboard-controller/blob/1.22.4/src/animated.tsx#L154-L190) creates two handlers and supplies [] to both.
2. [reanimated.native.ts](https://github.com/kirillzyusko/react-native-keyboard-controller/blob/1.22.4/src/reanimated.native.ts#L13-L17) forwards that array to useHandler.
3. Reanimated’s native implementation emits the warning.

So it appears to be a compatibility warning introduced by Reanimated 4.6 rather than a functional bug in keyboard-controller. The proposed change only stops forwarding the dependency argument in the .native implementation; the public signature and Web implementation remain unchanged.

kirillzyusko:
@huextrat okay, makes sense! Thanks for explanation!

Could you please also remove unit test? What about `useGenericKeyboardHandler` and `useFocusedInputHandler` hooks? Should we omit `deps` from `useHandler` there as well?

huextrat:
Good catch! I removed the unit test and updated both `useGenericKeyboardHandler` and `useFocusedInputHandler` as well. They now forward `deps` only on Web, so the native implementation receives `undefined` and doesn't emit the Reanimated 4.6 warning while preserving the existing Web behavior. Thanks!

kirillzyusko:
@huextrat I think it's fine to omit deps at all for now (without conditional code) 🙃 

This hook doesn't work on web anyway and I don't think I can add web support anytime soon 🤷‍♂️ 

Or maybe we need to create our version of `useHandler`, which will skip deps on native side and will pass them on web. I def wouldn't like to have such platfform specific checks directly in the code

huextrat:
Makes sense. I've removed the platform check and now omit `deps` from `useHandler` in both hooks. The dependency arrays are still used by `useLayoutEffect`. Pushed, thanks!

kirillzyusko:
Hey @huextrat 

Thank you for quick fixes. I just want to ask your opinion 👀 

So effectively `useHandler` just returns `doDependenciesDiffer`/`context` variables. It can be derived using babel plugin, so we don't need to pass deps array.

On web there is no such concept as worklet, so people may not use babel plugin, in this case we need to pass `deps`.

>  I don't think I can add web support anytime soon 🤷‍♂️

I checked actual state, and some important PRs were merged few weak ago in WebKit, so maybe in iOS 27 the web support can be landed too 🤷‍♂️ 

So I'm just thinking - would it make sense to make `useEvent` handler to be platform dependent? We already have `reanimated.native.ts`/`reanimated.ts` files, so web version can just export original `useEvent` hook and `native` can export own version without passing the `deps` array?

In this case we would need to swap imports? What do you think about it? I think platform-dependent implementation should be more effective here?

---

I just want to make changes in such way, so that when I will implement web I will not need to move code around - just want to produce as less changes as possible.

huextrat:
Thanks for the suggestion! I've updated the implementation accordingly in `123e66f`.

`useHandler` now lives behind the existing platform-specific `reanimated.native.ts` / `reanimated.ts` modules. The native wrapper deliberately calls Reanimated without the dependency array, while the Web module re-exports the original hook and keeps receiving dependencies. `useGenericKeyboardHandler` and `useFocusedInputHandler` now import this platform-specific helper, so there are no inline platform checks and the Web behavior is preserved for future support.

I interpreted the `useEvent` mention as referring to `useHandler`, since `useHandler` is the hook that receives the dependency array. Please let me know if you had a separate `useEvent` abstraction in mind.

kirillzyusko:
Thank you! Let's wait for green CI and we can merge it! ✅

kirillzyusko (approved):

PR #1624, opened 2026-09-04
docs: mention missing animations in `KeyboardStickyView`/`KeyboardToolbar` and the recommended fix

kirillzyusko:
## 📜 Description

Recommend to enable `DISABLE_COMMIT_PAUSING_MECHANISM` feature flag in reanimated.

## 💡 Motivation and Context

`KeyboardStickyView` has been migrated to reanimated recently.

At the moment reanimated has a bug, when animation can be canceleld because react will perform another commit.

In this PR I recommend to enable specific feature flags to avoid this probelm. We recommended to turn this flag on for `KeyboardAvoidingView` and for `KeyboardChatScrollView`.

### Docs

- recommend to enable `DISABLE_COMMIT_PAUSING_MECHANISM` if you are using `KeyboardStickyView`/`KeyboardToolbar` + new arch.

Tested via deployed version.

image

## 📝 Checklist

- [x] CI successfully passed

### DIFF
--- a/docs/docs/api/components/keyboard-sticky-view/index.mdx
+++ b/docs/docs/api/components/keyboard-sticky-view/index.mdx
@@ -69,3 +69,21 @@ An object containing next properties:
 
 - **closed** - additional offset to the view when keyboard is closed. Default value is `0`.
 - **opened** - additional offset to the view when keyboard is opened. Default value is `0`.
+
+## Troubleshooting
+
+### Missing animations on iOS (new arch only)
+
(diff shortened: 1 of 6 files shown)


sommeeeer:
@kirillzyusko Does this cause the animation to not happen at all? Like it would apply 0 padding so the keyboard is actually not pushed upwards?

kirillzyusko:
@sommeeeer no, the animation will be just instant in this case, i. e. it will shift content by `300px`, but will do it instantly without any animations 🤷‍♂️

sommeeeer:
Alright :)

PR #1641, opened 2026-09-15
fix: clamp `KeyboardChatScrollView` scroll target at 0 on shrink

### LINKED ISSUE #1636: KeyboardChatScrollView: shrinking extraContentPadding sets a negative contentOffset on iOS (blank gap above short content)
### Describe the bug

In `KeyboardChatScrollView` (non-inverted), when `extraContentPadding` **shrinks** while the content is shorter than the viewport, `useExtraContentPadding` computes a **negative** scroll target and applies it. On iOS/Fabric that target is written straight into `contentOffset`, so the list gets pushed down and a large blank gap appears above the content.

The scroll target is clamped at the top (`maxScroll`) but not at `0`:

https://github.com/kirillzyusko/react-native-keyboard-controller/blob/main/src/components/KeyboardChatScrollView/useExtraContentPadding/index.ts#L136-L146

```ts
const maxScroll = Math.max(
  size.value.height - layout.value.height + currentTotal,
  0,
);
const target = Math.min(scroll.value + effectiveDelta, maxScroll); // can be < 0

scrollToTarget(target);
```

With a short (non-scrollable) list `scroll.value` is `0` and `isScrollAtEnd()` is `true`, so `shouldShiftContent()` passes; when the composer collapses, `effectiveDelta` is negative (e.g. `-80`) and `target` becomes `-80`. `scrollToTarget` then does `contentOffsetY.value = -80` on Fabric, which sticks (it is a set `contentOffset`, not a bounce), leaving a ~80px empty band at the top of the list.

The `inverted` branch right above it does clamp its lower bound (`Math.max(..., -currentTotal)`), so this looks like an oversight rather than intended behaviour.

### Steps to reproduce

1. `KeyboardChatScrollView` with `keyboardLiftBehavior="whenAtEnd"` and a multiline `TextInput` inside a `KeyboardStickyView`, wired via `extraContentPadding` (composer growth is animated with `withTiming`).
2. Start with an **empty / very short** list, so the content does not fill the viewport.
3. Focus the input and type until it grows to several lines.
4. Submit — clear the text so the composer collapses back to one line (`extraContentPadding` animates back to `0`).

### Expected

Content stays flush with the top of the scroll view.

### Actual

The content is pushed down by roughly the composer's collapsed height; the taller the composer was, the wider the blank gap.

### Suggested fix

Clamp the non-inverted target at `0`:

```ts
const target = Math.max(
  Math.min(scroll.value + effectiveDelta, maxScroll),
  0,
);
```

### Environment

- react-native-keyboard-controller: 1.22.4
- react-native: 0.85.2, New Architecture (Fabric) enabled
- Platform: iOS (Android unaffected — it goes through the deferred `scrollTo` path, which the native ScrollView clamps)

HuuNguyen312:
## 📜 Description

`useExtraContentPadding` clamps the non-inverted scroll target only at the top (`maxScroll`), not at `0`:

```ts
const target = Math.min(scroll.value + effectiveDelta, maxScroll); // can be < 0
```

With a list shorter than the viewport, `scroll.value` is `0`; when `extraContentPadding` shrinks, `effectiveDelta` is negative and the target goes negative. On Fabric iOS `scrollToTarget` writes it straight into `contentOffset`, so it sticks and a blank band appears above the content. The `inverted` branch right above already clamps its lower bound (`Math.max(..., -currentTotal)`).

This PR clamps the non-inverted target at `0` as well.

## 💡 Motivation and Context

Fixes #1636.

Real-world repro: a comment composer built with `KeyboardChatScrollView` (`keyboardLiftBehavior="whenAtEnd"`) plus a multiline `TextInput` in a `KeyboardStickyView` wired through `extraContentPadding`. On a short/empty thread, growing the composer to several lines and then sending (composer collapses back to one line) pushed the list down by roughly the composer's extra height.

### JS

- `useExtraContentPadding`: clamp the non-inverted scroll target at `0` so a shrinking `extraContentPadding` can no longer produce a negative `contentOffset`.

- Two regression tests added to `useExtraContentPadding/__tests__/contentOffsetY.spec.ts` covering both output paths (Fabric `contentOffsetY`, and the `scrollTo` fallback). Both fail on `main` (`-80`) and pass with the fix.
- `yarn test` — all `src/` suites pass (219 tests). The `example/`, `FabricExample/` and `docs/` suites fail on my machine both with and without this change because the example apps' dependencies aren't installed (`yarn bootstrap`); unrelated to this diff.
- `yarn typescript` — clean.
- `yarn lint` — clean.
- Verified on device (iOS 18, RN 0.85.2, New Arch) via a `patch-package` patch of 1.22.4 carrying this exact change: the blank gap is gone and normal lift/shrink behaviour on longer lists is unchanged.

Not provided — the change is a one-line clamp; the failing/passing unit tests above capture the behaviour.

## 📝 Checklist

- [x] CI successfully passed


kirillzyusko:
Hey @HuuNguyen312 

Would it be possible to add small code snippet that shows the problem (minimal one) and add before/after videos showing that the problem has been actually fixed? 🙏

kirillzyusko:
Hey @HuuNguyen312 

Just a friendly ping - could you prepare a minimal repro for these changes? So that I can run it and be sure that it's indeed a correct fix for your problem? 👀

